PR #9, opened 2024-02-28
Add a GitHub Actions main workflow and fix the build

### LINKED ISSUE #12: [BUG] Trait Bound Error in Stylo
### Description:
While working with Stylo, I encountered a trait bound error. The error message indicates that the trait `serde::Serialize` is not implemented for `Url`, which is required by `servo_arc::Arc<Url>: serde::Serialize`.

### Error Details:
The error occurs in the file `~\.cargo\git\checkouts\stylo-d5871ad1ba1940d3\e2a4834\style\servo\url.rs` at line 211. Here is the relevant code and error message:

```rust
211 | #[derive(Clone, Debug, Deserialize, MallocSizeOf, PartialEq, Serialize)]
    |                                                              ^^^^^^^^^ the trait `serde::Serialize` is not implemented for `Url`, which is required by `servo_arc::Arc<Url>: serde::Serialize`
...
216 |     Valid(#[ignore_malloc_size_of = "Arc"] Arc<Url>),
    |           ----------------------------------------- required by a bound introduced by this call
```

### Expected Behavior:
The `Url` type should be serializable as it is used within an `Arc` in the `Valid` variant of an enum (or similar data structure).

### Actual Behavior:
The compiler throws a trait bound error because `Url` does not implement `serde::Serialize`.

mrobinson:
Fixes #12.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,51 @@
+name: CI
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+  workflow_dispatch:
+  merge_group:
+    types: [checks_requested]
+
+
+jobs:
+  linux-debug:
+    name: Linux (Debug)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run Tests
+        run: cargo build --features servo
+        env:
+          RUST_BACKTRACE: 1
+
+  linux-release:
+    name: Linux (Release)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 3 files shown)


delan (commented):
Let’s merge this once we sort out the main branch situation, I’ll have a patch for that up soon. In the meantime, don’t forget to check in the workflow.yml :)

mrobinson:
Whoops. This should be ready to go now. Checks should be showing up as passing below.

.github/workflows/main.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run Tests
+        run: cargo build --release --features servo
+        env:
+          RUST_BACKTRACE: 1
+
+  build_result:
```
mukilan, line 36: Nit: the other jobs use the dash as a separator but here we use the underscore.

mukilan (approved):

PR #10, opened 2024-02-29
Enable `box-shadow` for Layout 2020

mrobinson:


Loirooriol (approved):

Loirooriol:
BTW, I guess It would be better for PRs to use branches from forks, instead of adding branches here in the main repo.

mrobinson:
> BTW, I guess It would be better for PRs to use branches from forks, instead of adding branches here in the main repo.

Yeah. That makes total sense.

PR #11, opened 2024-03-01
Mirror the `main` branch to the release branch

mrobinson:


delan (changes requested):
This is a good start, but it relies on us remembering to update the `dest`. Can we detect the date automatically with something like this in a shell step?

```sh
first_commit=$(git log --pretty=\%H --grep='Servo initial downstream commit')
upstream_base=$first_commit~
dest=$(git log -n1 --pretty='%as' "$upstream_base")
```

What happens (or should happen) when the push is not a fast forward?

mrobinson:
@delan  Okay, I''ve updated this with the changes requested. Here's an example of the action run against a non-fast forward push:

https://github.com/mrobinson/stylo/actions/runs/8139727004/job/22243528272

delan (approved):
Thanks!

PR #13, opened 2024-03-04
Upgrade to WebRender 0.64

mrobinson:


mrobinson:
I've converted this to a draft until the corresponding change in Servo is ready. I thought this would be easier than it has turned out to be. :sweat_smile:

Loirooriol (approved):

PR #14, opened 2024-03-04
Remove Servo-specific font code and expose font-face properties

mrobinson:
Servo needs these properties in order to properly load the fonts on
servo.org. In addition, remove some custom Servo code that is no longer
necessary. Servo will handle filtering web font sources internally,
which will allow it to be less dependent on Stylo for this.

This is the first step toward WOFF2 support in Servo.


mrobinson:
The filtering code is part of the companion Servo PR at: https://github.com/servo/servo/pull/31601

jdm (approved):

PR #15, opened 2024-03-05
Update length_percentage.rs

sepcnt:
Add resolve_percentage public method, resolves to a <percentage> with given basis.
When utilizing stylo into other renderers, it will be helpful to convert LengthPercentage to ratio.

### DIFF
--- a/style/values/computed/length_percentage.rs
+++ b/style/values/computed/length_percentage.rs
@@ -442,6 +442,20 @@ impl LengthPercentage {
         }
     }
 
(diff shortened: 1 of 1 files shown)


mrobinson (commented):
This sort of change, one affecting Gecko and Servo alike and dealing with the basics of stylo, should be contributed to upstream stylo instead of our downstream fork. We are currently in the process of syncing our version of stylo with Gecko's so are reducing our diff.

PR #19, opened 2024-03-16
Update Cargo.lock after the changes from #13

Loirooriol:

PR #20, opened 2024-03-16
Use --locked when building in CI

Loirooriol:
Thus the build will fail if a PR modifies Cargo.toml but forgets to include the changes in Cargo.lock


Loirooriol:
See https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options

Loirooriol:
Works as expected:
- CI failed brfore landing #19: https://github.com/servo/stylo/actions/runs/8308496963
- CI passes after landing #19: https://github.com/servo/stylo/actions/runs/8309802096

jdm (approved):

PR #21, opened 2024-03-18
Add support for `start`, `end`, and `space-evenly` values to `align-` and `justify-content` properties

nicoburns:
Adds the `start`, `end` and `space-evenly` values to the `align-content` and `justify-content` enums. I will post a Servo PR implementing support for these values in layout2020's flexbox implementation shortly.

Corresponding servo PR: https://github.com/servo/servo/pull/31724

I have made the simplest change and just added these to Servo's existing enum, but the change I would really like to make is to switch Servo over to [Gecko's content alignment implementation](https://github.com/servo/stylo/blob/main/style/values/specified/align.rs). This adds support for parsing:
   - ~`auto` and~ `normal`
   - `baseline`, `first baseline`, and `last baseline`
   - ~`self-start` and `self-end`~
   - The "safe" alignment variants that fallback behaviour on overflow
   
EDIT: While the `self-start` and `self-end` values are representable by the underlying `AlignFlags` type, they are not parsed by the `ContentDistribution` wrapper.
   
Servo doesn't support all of those yet, but it also doesn't support `stretch` but still parses it, and I feel like it would be good to try to reduce the differences between the Servo and Gecko builds of Stylo.

P.S. Should I also be submitting this upstream to Mozilla (presumably only once it's accepted here?)? Or is there some kind of sync process that does this automatically?

### DIFF
--- a/style/properties/longhands/position.mako.rs
+++ b/style/properties/longhands/position.mako.rs
@@ -97,7 +97,7 @@ ${helpers.single_keyword(
     // FIXME: Update Servo to support the same Syntax as Gecko.
     ${helpers.single_keyword(
         "justify-content",
-        "flex-start stretch flex-end center space-between space-around space-evenly",
+        "start flex-start stretch end flex-end center space-between space-around space-evenly",
         engines="servo",
         servo_pref="layout.flexbox.enabled",
         extra_prefixes="webkit",
@@ -137,7 +137,7 @@ ${helpers.single_keyword(
     // FIXME: Update Servo to support the same Syntax as Gecko.
     ${helpers.single_keyword(
         "align-content",
-        "stretch flex-start flex-end center space-between space-around space-evenly",
+        "stretch start flex-start end flex-end center space-between space-around space-evenly",
         engines="servo",
         servo_pref="layout.flexbox.enabled",
         extra_prefixes="webkit",
--- a/style/properties/longhands/position.mako.rs
+++ b/style/properties/longhands/position.mako.rs
@@ -97,7 +97,7 @@ ${helpers.single_keyword(
     // FIXME: Update Servo to support the same Syntax as Gecko.
     ${helpers.single_keyword(
         "justify-content",
-        "flex-start stretch flex-end center space-between space-around",
+        "flex-start stretch flex-end center space-between space-around space-evenly",
         engines="servo",
         servo_pref="layout.flexbox.enabled",
         extra_prefixes="webkit",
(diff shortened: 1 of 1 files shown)


style/properties/longhands/position.mako.rs
```diff
@@ -95,7 +95,7 @@ ${helpers.single_keyword(
     // FIXME: Update Servo to support the same Syntax as Gecko.
     ${helpers.single_keyword(
         "justify-content",
-        "flex-start stretch flex-end center space-between space-around",
```
mrobinson, line 98: Is it possible for Servo to share the version of this for Gecko that's used below now?

style/properties/longhands/position.mako.rs
```diff
@@ -132,7 +132,7 @@ ${helpers.single_keyword(
     // FIXME: Update Servo to support the same Syntax as Gecko.
     ${helpers.single_keyword(
         "align-content",
-        "stretch flex-start flex-end center space-between space-around",
```
mrobinson, line 135: I have the same question here essentially/

style/properties/longhands/position.mako.rs
nicoburns: I discuss this in the PR description. It absolutely is, but that will mean that we're parsing values that we don't actually support (but we already are (`stretch`), and are all over the place when it comes to Flexbox styles).

mrobinson: I guess the main question here is do you plan to implement support for these values in the mid-term. I think that if you do plan to add support for those values, it is worth just unifying the implementation now.

mrobinson:
Regarding the question from the PR description:

> P.S. Should I also be submitting this upstream to Mozilla (presumably only once it's accepted here?)? Or is there some kind of sync process that does this automatically?

In this case the right repository to submit this to is this one. This is the sort of change we plan to upstream to Gecko once we get into sync. The reasoning is that we need this change for Servo.

style/properties/longhands/position.mako.rs
nicoburns: I think the answer is largely a yes. `normal` is just the default value. The "safe" variants should be pretty trivial to implement (it's just "fallback to `start` on overflow"). Baseline alignment is also very much doable, although tricky to do performantly.

(and on a closer reading of the Gecko implementation the other values I listed aren't relevant - they are there because Gecko is sharing an underling bitflags type between the content alignment and item alignment types but they will never be returned by the parser for these properties).

nicoburns:
@mrobinson I've rebased this on top of the latest main. I think we probably will want to switch to the Gecko implementation, and I am able to work on that if there is consensus. But my preference would be to land this first and then proceed with that as a separate change. Partly because it is already implemented (and it's such a tiny change), and partly because that allows us to separate algorithmic changes from type refactoring on the Servo side, and there may well be some other properties we want to switch over at the same time (e.g. `align-items`/`justify-items`).

P.S. How does synchronising landing changes here and in the servo repo work? Land here first and then update the Servo PR to point to the new commit? (Is Servo relying on `Cargo.lock` to pin a specific Stylo commit? - and is there a reason why it's not pinning a specific commit in Cargo.toml?)

Loirooriol (approved):

PR #24, opened 2024-03-19
Enable x-lang attribute for Servo engine

MunishMummadi:
### Description

This Pull Request includes the following changes:

#### `style/properties/longhands/font.mako.rs`
- Modified the `helpers.predefined_type` function call for `-x-lang`, adding support for `gecko servo` engines.

#### `style/values/specified/font.rs`
- Added implementation for `XTextScale`, providing functionality for handling text scaling.

refer at [servo issue](https://github.com/servo/servo/issues/31723)

### DIFF
--- a/style/values/specified/font.rs
+++ b/style/values/specified/font.rs
@@ -2007,6 +2007,8 @@ impl XTextScale {
     ToCss,
     ToResolvedValue,
     ToShmem,
+    Serialize,
+    Deserialize,
 )]
 /// Internal property that reflects the lang attribute
 pub struct XLang(#[css(skip)] pub Atom);
--- a/style/values/specified/font.rs
+++ b/style/values/specified/font.rs
@@ -2007,9 +2007,8 @@ impl XTextScale {
     ToCss,
     ToResolvedValue,
     ToShmem,
-    Serialize,
-    Deserialize,
 )]
+#[cfg_attr(feature = "servo", derive(Deserialize, Serialize))]
 /// Internal property that reflects the lang attribute
 pub struct XLang(#[css(skip)] pub Atom);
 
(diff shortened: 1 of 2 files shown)


Loirooriol:
The patch looks good, but here in the description you are mentioning `XTextScale` which is unrelated?

style/values/specified/font.rs
```diff
@@ -2007,6 +2007,8 @@ impl XTextScale {
     ToCss,
     ToResolvedValue,
     ToShmem,
+    Serialize,
+    Deserialize,
```
Loirooriol, line 2011: Nit: not a big deal because this change won't get into Firefox, but I think it's better to preserve compatibility and use this instead:

```rs
#[derive(
    Clone,
    Debug,
    MallocSizeOf,
    PartialEq,
    SpecifiedValueInfo,
    ToComputedValue,
    ToCss,
    ToResolvedValue,
    ToShmem,
)]
#[cfg_attr(feature = "servo", derive(Deserialize, Serialize))]
/// Internal property that reflects the lang attribute
pub struct XLang(#[css(skip)] pub Atom);
```

This will add `Deserialize, Serialize` just for Servo, not for Gecko.

Loirooriol (changes requested):

style/values/specified/font.rs
emilio: It's fine to make stuff serializable in Gecko generally, fwiw.

emilio: (Unless Atom is somehow not serializable in Firefox for some reason)

Loirooriol: Yeah Atom not serializable in Firefox:
```
 0:16.17 error[E0277]: the trait bound `Atom: Serialize` is not satisfied
 0:16.18     --> servo/components/style/values/specified/font.rs:2006:5
 0:16.18      |
 0:16.18 2006 |     Serialize,
 0:16.18      |     ^^^^^^^^^ the trait `Serialize` is not implemented for `Atom`
 0:16.18 ...
 0:16.18 2010 | pub struct XLang(#[css(skip)] pub Atom);
 0:16.18      |                  --------------------- required by a bound introduced by this call
```

PR #34, opened 2024-04-23
Switch to `markup5ever`'s version of `Atom` and upgrade `xml5ever`

mrobinson:
`html5ever` was just rexporting `markup5ever`'s version of Atom's so it
makes sense to loosen the dependencies and switch to the ones in
`markup5ever`. This should make it easier to keep things up to date in
the future.

In addition, we are now using the latest version of `xml5ever`.


mrobinson:
@Loirooriol This can wait until after the style upgrade if that makes things easier for you.

nicoburns:
This makes a lot of sense. Another thing I was going to suggest along similar lines is merging `servo_atoms` and the atoms in `markup5ever` into a combined `html_atoms` crate or similar (or just into `markup5ever`). Excepting a few `-moz` properties (which could be feature flagged) there isn't really anything servo-specific about `servo_atoms`, and it would nice to gradually remove `stylo`'s dependencies on `servo_*` crates where possible.

Loirooriol:
I rebased the patch but now it seems to be failing the compilation, even though it works locally...
Hopefully I didn't break anything with the style upgrade.

PR #35, opened 2024-04-25
Test

Loirooriol:
Just to check if the checks pass on main


Loirooriol:
They fail :( https://github.com/servo/stylo/actions/runs/8834658458/

mrobinson:
This might be due to a change the GitHub Runner images. It looks like it is failing to find the `mako` Python module.

Loirooriol:
I think it's actually https://phabricator.services.mozilla.com/D204532, I will try reverting it

PR #37, opened 2024-04-25
Turn white-space into a shorthand

Loirooriol:
Split it into `white-space-collapse` and `text-wrap-mode`:

| white-space | white-space-collapse | text-wrap-mode |
| ----------- | -------------------- | -------------- |
| normal      | collapse             | wrap           |
| nowrap      | collapse             | nowrap         |
| pre-wrap    | preserve             | wrap           |
| pre         | preserve             | nowrap         |
| pre-line    | preserve-breaks      | wrap           |
| -           | preserve-breaks      | nowrap         |

Note this introduces a combination that wasn't previously possible.

### DIFF
--- a/style/properties/longhands/inherited_text.mako.rs
+++ b/style/properties/longhands/inherited_text.mako.rs
@@ -145,58 +145,17 @@ ${helpers.predefined_type(
     affects="layout",
 )}
 
-<%helpers:single_keyword
-    name="white-space"
-    values="normal pre nowrap pre-wrap pre-line"
-    engines="servo",
-    gecko_enum_prefix="StyleWhiteSpace"
-    needs_conversion="True"
-    animation_value_type="discrete"
-    spec="https://drafts.csswg.org/css-text/#propdef-white-space"
-    servo_restyle_damage="rebuild_and_reflow"
-    affects="layout"
->
-    impl SpecifiedValue {
-        pub fn allow_wrap(&self) -> bool {
-            match *self {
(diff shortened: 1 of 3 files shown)


style/properties/longhands/inherited_text.mako.rs
```diff
@@ -145,55 +145,12 @@ ${helpers.predefined_type(
...
 
-<%helpers:single_keyword
-    name="white-space"
-    values="normal pre nowrap pre-wrap pre-line"
-    engines="servo",
-    gecko_enum_prefix="StyleWhiteSpace"
-    needs_conversion="True"
-    animation_value_type="discrete"
-    spec="https://drafts.csswg.org/css-text/#propdef-white-space"
-    servo_restyle_damage="rebuild_and_reflow"
```
Loirooriol, line 156: I'm not sure if `servo_restyle_damage` or `needs_conversion` need to be kept.

mrobinson: Yes, let's keep these for now. They are used for incremental layout. It's unclear how much of these damage flags we will keep for Layout 2020, but I think we shouldn't modify them as much as possible until we have incremental layout in place.

mrobinson (approved):
I'm very happy to see this. It greatly reduces the differences we have from Gecko.

style/properties/longhands/inherited_text.mako.rs
Loirooriol: I have added these to both longhands.

PR #47, opened 2024-06-20
Blitz tracking PR

nicoburns:
Putting this up to track the Stylo changes we have made for Blitz and that we would like to see upstreamed into Servo's version in this repo. Servo will also want these in it's own right as it comes to implement a full implementation of the CSS Grid, `gap`, and alignment styles.

~~Not necessarily expecting this to merged as-is~~ (UPDATE: given improvements to the feature flagging, I feel like this probably could be merged / upstreamed now).

Corresponding Servo PR: https://github.com/servo/servo/pull/32686

### Upstream submissions

- [x] https://bugzilla.mozilla.org/show_bug.cgi?id=1906395
- [x]  https://bugzilla.mozilla.org/show_bug.cgi?id=1906994
- [x] https://bugzilla.mozilla.org/show_bug.cgi?id=1906995

### Changes made

- Built on top of https://github.com/servo/stylo/pull/51
- Cherry-picked https://github.com/servo/stylo/commit/ddb7af7127e6c6c64e6202fab00d42aa4b33a5c3 from upstream
- Remove alignment style adjustment function (which should be implemented in layout anyway - implemented in https://github.com/servo/servo/pull/32686)
- Add / make public a couple of helper methods to AlignFlags
- Switch to Gecko's representation of alignment styles
- Enable `row-gap` and the shorthand `gap` (behind flexbox flag)
- Switch `column-gap` to being behind the flexbox flag rather than the multi-column layout flag
- Enable CSS Grid styles (styles shared by flexbox and grid are behind the flexbox flag, so you need to enable both for grid support)
- Enable aspect_ratio in layout_2020 (this is now behind the flexbox flag rather than the legacy-layout flag)

### Notes

- I think the "use gecko alignment implementation" changes probably could be upstreamed as-is (but this will require a corresponding servo PR)
- The grid-only styles are also quite straightforward because they're behind a flag.
- The case where it's more tricky is the styles that are shared by Flexbox and Grid because Stylo currently only allows styles to be gated behind a single pref. There are a few options here:
   1. We could just make them enabled unconditionally on the basis that implementation are coming soon.
   2. We could update stylo to accept a list of prefs. This would be quite easy to implement but would have a small additional runtime cost.
   3. We could add separate "alignment" and "gap" preferences that need to be enabled seperately. This would be simple but would be cumbersome when trying to actually use them.
   4. We could do the separate prefs as above + add some mechanism to automatically enable prefs
   
   I'd probably favour (ii).
   
 Currently based on: 2024-05-31


style/properties/longhands/position.mako.rs
```diff
@@ -93,33 +92,19 @@ ${helpers.single_keyword(
...
+    "justify-content",
+    "JustifyContent",
+    "specified::JustifyContent(specified::ContentDistribution::normal())",
+    engines="gecko servo",
+    spec="https://drafts.csswg.org/css-align/#propdef-justify-content",
+    extra_prefixes="webkit",
+    animation_value_type="discrete",
+    servo_restyle_damage="reflow",
+    affects="layout",
+)}
```
mrobinson, line 105: It's great to unite these, but we shouldn't lose the `servo_pref` setting here in the process.

style/properties/longhands/position.mako.rs
```diff
@@ -133,45 +118,45 @@ ${helpers.single_keyword(
...
+    "align-content",
+    "AlignContent",
+    "specified::AlignContent(specified::ContentDistribution::normal())",
+    engines="gecko servo",
+    spec="https://drafts.csswg.org/css-align/#propdef-align-content",
+    extra_prefixes="webkit",
+    animation_value_type="discrete",
+    servo_restyle_damage="reflow",
+    affects="layout",
+)}
```
mrobinson, line 131: This also needs the pref that has been lost.

style/properties/longhands/position.mako.rs
```diff
@@ -133,45 +118,45 @@ ${helpers.single_keyword(
...
+    "align-items",
+    "AlignItems",
+    "specified::AlignItems::normal()",
+    engines="gecko servo",
+    spec="https://drafts.csswg.org/css-align/#propdef-align-items",
+    extra_prefixes="webkit",
+    animation_value_type="discrete",
+    servo_restyle_damage="reflow",
+    affects="layout",
+)}
```
mrobinson, line 143: Ditto.

style/properties/longhands/position.mako.rs
```diff
@@ -133,45 +118,45 @@ ${helpers.single_keyword(
...
+impl_align_conversions!(crate::values::specified::align::AlignItems);
```
mrobinson, line 144: Why is this added?

style/properties/longhands/position.mako.rs
```diff
@@ -183,34 +168,6 @@ ${helpers.single_keyword(
...
-        "JustifyItems",
-        "computed::JustifyItems::legacy()",
-        engines="gecko",
-        spec="https://drafts.csswg.org/css-align/#propdef-justify-items",
-        animation_value_type="discrete",
-        affects="layout",
-    )}
-
-    #[cfg(feature = "gecko")]
-    impl_align_conversions!(crate::values::specified::align::JustifyItems);
```
mrobinson, line 213: I think it's better not to adjust the position of these, as this makes the diff with upstream harder to process.

style/properties/longhands/position.mako.rs
```diff
@@ -240,46 +197,26 @@ ${helpers.predefined_type(
...
+${helpers.predefined_type(
+    "justify-self",
+    "JustifySelf",
+    "specified::JustifySelf(specified::SelfAlignment::auto())",
+    engines="gecko servo",
+    spec="https://drafts.csswg.org/css-align/#justify-self-property",
+    animation_value_type="discrete",
+    affects="layout",
+)}
+impl_align_conversions!(crate::values::specified::align::SelfAlignment);
```
mrobinson, line 219: You've lost the servo_pref again here.

style/properties/longhands/position.mako.rs
```diff
@@ -476,7 +417,7 @@ ${helpers.predefined_type(
     "row-gap",
     "length::NonNegativeLengthPercentageOrNormal",
     "computed::length::NonNegativeLengthPercentageOrNormal::normal()",
-    engines="gecko",
+    engines="gecko servo",
```
mrobinson, line 420: I think this needs the new `servo_pref`.

style/properties/shorthands/position.mako.rs
```diff
@@ -128,7 +128,7 @@
 
 <%helpers:shorthand
     name="gap"
-    engines="gecko"
+    engines="gecko servo"
```
mrobinson, line 131: Ditto.

style/properties/shorthands/position.mako.rs
```diff
@@ -716,7 +720,7 @@
 
 <%helpers:shorthand
     name="place-content"
-    engines="gecko"
+    engines="gecko servo"
```
mrobinson, line 723: For all of these as well, I think.

style/style_adjuster.rs
```diff
@@ -402,29 +402,6 @@ impl<'a, 'b: 'a> StyleAdjuster<'a, 'b> {
...
-                AlignItems::Stretch => AlignSelf::Stretch,
-                AlignItems::Baseline => AlignSelf::Baseline,
-                AlignItems::FlexStart => AlignSelf::FlexStart,
-                AlignItems::FlexEnd => AlignSelf::FlexEnd,
-                AlignItems::Center => AlignSelf::Center,
-            };
-            self.style.mutate_position().set_align_self(self_align);
-        }
-    }
-
```
mrobinson, line 427: Nice.

mrobinson (changes requested):
This is a nice cleanup. Is there a corresponding Servo change. It would be nice to have that. In addition, I think it might make more sense to contribute this directly to upstream stylo and then wait until the next sync, to avoid too much of a hassle when we update. We need to start doing that, I think.

style/properties/longhands/position.mako.rs
nicoburns: Sure. Blitz will just enable both, so that works for us. I can go through and update the flags to use the flexbox flag for the case where it is needed for both. Although it wouldn't be that hard to add support for multiple flags if that was desirable (just swap out the `Option<&str>` for `&[&str]` (or `Option<&[&str]>`)

style/properties/longhands/position.mako.rs
nicoburns: It was already present in the Gecko version, I just made it unconditional (this doesn't show up very well in the diff due to unnesting). It adds conversions to and from u8. I'm not entirely sure why that's useful, but it didn't seem particularly gecko-specific. I could make it conditional again if that's preferred.

nicoburns:
Upgraded to 2024-05-31

style/properties/longhands/position.mako.rs
nicoburns: I think I've put them back in the order they were in.

style/properties/shorthands/position.mako.rs
```diff
@@ -808,11 +815,37 @@
             Ok(())
         }
     }
+
+    #[cfg(feature = "servo")]
```
nicoburns, line 819: We need a different impl here because we're feature flagging these properties and gecko isn't.

nicoburns:
I've cherry-picked [ddb7af7127e6c6c64e6202fab00d42aa4b33a5c3](https://github.com/servo/stylo/commit/ddb7af7127e6c6c64e6202fab00d42aa4b33a5c3) from upstream, which along with https://github.com/servo/stylo/pull/51 makes the diff (both against `main` and `upstream` a lot cleaner).

nicoburns:
All changes from this PR have now been upstreamed and synced back to this repo :) Closing this. If we need further changes in future I can open a new PR.

PR #52, opened 2024-07-11
Correct repository link

mpalmer:
(It might also be worth bumping `version` to match what's published on crates.io)


mrobinson:
The best place to apply this fix is upstream: https://searchfox.org/mozilla-central/source/servo/components/selectors/Cargo.toml. This is a downstream repository which maintains a set of patches on top of upstream.

PR #58, opened 2024-07-29
Reinstate mako PATH fix

### LINKED ISSUE #57: Stylo build is broken when building in a non-servo context
This is the same issue as #55, but #55 only solved the issue for Stylo CI. Using Stylo's `main` branch as git dependency for example currently fails with "mako not found".

I am open to other solutions, but as a short-term fix to make `main` non-broken I would be inclined to simply re-add the one-line import that we used to have. I would note that the current state of `main` is that the vendored python "wheel" is still included in the source-tree. It is just unused as the import has been removed.

The line we would need to reinstate is https://github.com/servo/stylo/commit/4de6af12dd74f9a8808cd5f2ffe34b33d368ee5d

The Gecko patch that removed it https://phabricator.services.mozilla.com/D204532 which also added the entire mako source code in another part of the tree.

nicoburns:
Fixes #57

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,8 +17,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-      - name: Install mako
-        run: sudo apt install python3-mako
       - name: Run Tests
         run: cargo build --features servo
         env:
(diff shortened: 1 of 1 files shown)


mrobinson:
Instead of this workaround it would be better to understand why mako from the virtual environment isn't being picked up when compiling stylo.

nicoburns:
> Instead of this workaround it would be better to understand why mako from the virtual environment isn't being picked up when compiling stylo.

Which virtual environment? Stylo (without servo/gecko) doesn't set one up as far as I can see.

mrobinson (changes requested):
Looks good now that I understand what this is doing. Please also revert the changes to the CI configuration though.

nicoburns:
> Please also revert the changes to the CI configuration though.

Done!

mrobinson (approved):

PR #59, opened 2024-07-30
Add `securitypolicyviolation` to servo_atoms

chocolate-pie:
Add `securitypolicyviolation` to servo_atoms for initial support for [servo/servo#25182](https://redirect.github.com/servo/servo/issues/25182).


mrobinson:
It feels wrong that we are adding Servo-specific atoms to Stylo. Is there a way that Servo can use its own atoms, perhaps by moving Servo-specific atoms to a crate within Servo?

PR #62, opened 2024-08-06
Enable `clip-path` for layout 2020

chocolate-pie:
Enable `clip-path` for layout_2020 to fix a part of [servo/servo#30514](https://redirect.github.com/servo/servo/issues/30514).


mrobinson:
This seems fine, but it should wait until there's a corresponding Servo PR implementing this feature before landing. I'm curious how this will work though, as implementing it requires path rasterization, which we do not have support for in layout.

chocolate-pie:
Path rasterization seems difficult for me, so I would like to implement only `circle()`, `ellipse()` and `rect()` support. 
(I've attempted to support `polygon()` property but crashing with `bug: image-masks not expected on rect/quads`)
https://github.com/chocolate-pie/servo/tree/support-clip-path

nicoburns:
Path rasterisation seems like something it would be appropriate to pull in a library for. Although a smaller one would be better. Perhaps tiny-skia?

mrobinson:
> Path rasterisation seems like something it would be appropriate to pull in a library for. Although a smaller one would be better. Perhaps tiny-skia?

This would be handled by a library like skia or vello, which would rasterize clip paths into mask images which could be used by WebRender.

Handling just circular and ovular paths via rounded rect clips seems like a decent workaround though.

mrobinson:
> @mrobinson I've spent two weeks on this, but I could not figure out the proper way to use `define_clip_image_mask`. Is it possible to only support `ellipse()`, `circle()` and `rect()`?

Yes, I think that is preferable for now, as the conversation about what rasterization library is a tricky one and we should be careful to try to use the same one that we use for canvas -- ie have only a single rasterizer for paths in Servo.

mrobinson:
> Thanks, I'll implement blob rasterizer later on. Could you merge this PR? This PR is needed to implement initial support for `clip-path`.

We typically only merge PRs in Stylo until there's a corresponding PR ready to land in Servo. In the meantime, you can use the PR as the reference for all [Stylo crates in your branch](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) using `rev = "refs/pull/62/head"`.

mrobinson (approved):

PR #66, opened 2024-08-22
Sync Servo's pseudo-class list with the one upstream

mrobinson:
Accepting pseudo-classes that are not supported shouldn't really
happen according to the specification, but since Servo doesn't
have a lot of use in production, we do this until we add support for
them. This allows declarations that use a lot of pseudo-classes to
succeed, even if one isn't supported. This is quite common and happens
even on servo.org.


Loirooriol (approved):
Accepting pseudo-classes that are not supported shouldn't really happen according to the spec, but I guess it's fine since people aren't using Servo for real yet and we plan to add support for them.

PR #67, opened 2024-09-06
Add test-wait

Taym95:
This is needed for https://github.com/servo/servo/issues/33342

### DIFF
--- a/atoms/static_atoms.txt
+++ b/atoms/static_atoms.txt
@@ -113,6 +113,7 @@ ratechange
 readystatechange
 referrer
 reftest-wait
+test-wait
 rejectionhandled
 removetrack
 reset


jdm:
We can also use LocalName::from_str instead.

nicoburns (approved):
I feel like we probably ought to add `test-wait` here given that `reftest-wait` is already here. Also, I believe that adding extra static values to a `string_cache::Atom` is pretty cheap.

nicoburns:
@mrobinson It seems that I don't have the permissions for this repo that I have on other servo org repos (approve CI runs, have my reviews on PRs "count", merge PRs, etc). Is that a permissions misconfiguration or this repo meant to be special?

nicoburns:
> > I feel like we probably ought to add `test-wait` here given that `reftest-wait` is already here. Also, I believe that adding extra static values to a `string_cache::Atom` is pretty cheap.
> 
> In my opinion, we should move all of the Servo-specific Atoms to Servo and use a separate atom type for Stylo. We should probably do the same for markup5ever.

Hmm... I that would impose a conversion cost between the different atom types everytime you crossed from one crate to another? My suggestion would be to go the other way and publish a standalone `web_atoms` type that can be used across all the crates. A bonus would be that this could go on crates.io and wouldn't need to be part of the sync process.

Taym95:
We can close this PR you want, I did not knew about `LocalName::from_str` and I used it in other PR.

PR #71, opened 2024-09-11
Add Cargo.lock to gitignore

nicoburns:
We don't have `Cargo.lock` committed, so we should ignore it. Currently it keeps showing up as an untracked file which makes it hard to avoid accidentally committing it as part of unrelated changes. `.gitignore` is currently not upstream but is part of the "Initial downstream commit".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /_filtered/
 /target/
 /style/properties/__pycache__/
+Cargo.lock


Loirooriol:
Shouldn't we commit `Cargo.lock` instead?

mrobinson:
`Cargo.lock` used to be committed, but it increased the size of our diff from upstream. Ultimately, Stylo is used in Servo which has its own `Cargo.lock` file. As long as thing keep compiling here and in Servo, I think we do not have to commit `Cargo.lock`. It's a bit unusual for crates to check in their `Cargo.lock` file in any case.

nicoburns:
> It's a bit unusual for crates to check in their Cargo.lock file in any case.

I believe there was recently a change of advice from the Rust/Cargo teams about this. They previously recommended that libraries don't check in Cargo.toml, and they now weakly recommend doing so. But I think it's pretty much a toss up in terms of advantages/disadvantages, and many crates still don't. As it's a bit of a pain for Stylo I suggest we just don't for now (we can always commit it again later, but I would personally consider that low priority).

mrobinson (approved):
I think this is okay for now.

PR #75, opened 2024-09-13
Let `LengthPercentage::maybe_to_used_value()` accept `Option<Au>`

Loirooriol:
Given that `LengthPercentage::to_used_value()` accepts an `Au`, and `NonNegativeLengthPercentage::maybe_to_used_value()` an `Option<Au>`, it didn't make much sense for `LengthPercentage::maybe_to_used_value()` to accept an `Option<Length>`.

All callers from Servo and Stylo have an `Option<Au>` that they needed to convert into `Option<Length>` in order to call this function. Gecko doesn't call it.

### DIFF
--- a/style/values/computed/length_percentage.rs
+++ b/style/values/computed/length_percentage.rs
@@ -518,8 +518,8 @@ impl LengthPercentage {
 
     /// Convert the computed value into used value.
     #[inline]
-    pub fn maybe_to_used_value(&self, container_len: Option<Length>) -> Option<Au> {
-        self.maybe_percentage_relative_to(container_len)
+    pub fn maybe_to_used_value(&self, container_len: Option<Au>) -> Option<Au> {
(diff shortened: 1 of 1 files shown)


mrobinson:
@Loirooriol Nice! @atbrakhi and I were thinking the same thing about this. Do you mind opening up a companion revision in Mozilla's Phabricator and linking it here? That way everything will stay in sync.

Loirooriol:
Done in https://bugzilla.mozilla.org/show_bug.cgi?id=1918875 / https://phabricator.services.mozilla.com/D222210

mrobinson:
> Done in https://bugzilla.mozilla.org/show_bug.cgi?id=1918875 / https://phabricator.services.mozilla.com/D222210

Thank you very much!

mrobinson (approved):

PR #78, opened 2024-09-17
Fix stylo compilation on `wasm32-unknown-unknown` target

nicoburns:
Upstream bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1933015


nicoburns:
Submitted upstream as https://bugzilla.mozilla.org/show_bug.cgi?id=1933015

mrobinson:
Looks like this needs a bit of refinement since it was reverted upstream.

nicoburns:
This has landed upstream

PR #79, opened 2024-09-18
Enable min-content, max-content, fit-content and stretch

Loirooriol:
For the sizing properties.


mrobinson (approved):

nicoburns:
Should the remaining variants (`MozAvailable`, `WebkitFillAvailable`, and `FitContentFunction(_)`) also be enabled? I believe `MozAvailable` and `WebkitFillAvailable` are just non-standard aliases for `Stretch`, and `FitContentFunction(_)` shouldn't add much complexity on top of an implementation for `MinContent` and `MaxContent`.

Loirooriol:
We thought we would start with the basic ones, and think about others later.

PR #80, opened 2024-09-25
servo: Add support for parsing `:is()` and `:where()`

mrobinson:
Stylo supports these and they just need to be turned on. This change
also ensures that forgiving selector mode is turned off when parsing
selectors for the `@supports` rule.

### DIFF
--- a/style/servo/selector_parser.rs
+++ b/style/servo/selector_parser.rs
@@ -494,6 +494,31 @@ impl<'a, 'i> ::selectors::Parser<'i> for SelectorParser<'a> {
     type Impl = SelectorImpl;
     type Error = StyleParseErrorKind<'i>;
 
+    #[inline]
+    fn parse_nth_child_of(&self) -> bool {
+        false
+    }
+
+    #[inline]
+    fn parse_is_and_where(&self) -> bool {
+        true
+    }
+
+    #[inline]
+    fn parse_has(&self) -> bool {
+        false
+    }
+
+    #[inline]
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
Looks good. We've done this in Blitz and it seems to work. My one reservation would be that I believe that "nth-child" and "has" require extra invalidation logic that is not handled by Stylo and needs to be handled externally.

(I am keen to document how that works, probably as an example in the Stylo repo, but I haven't worked it out myself yet)

Loirooriol:
Yeah have you confirmed that `:has()` really works? Because I have strong doubts.

mrobinson:
> Yeah have you confirmed that `:has()` really works? Because I have strong doubts.

It's functional, but there's still work to do with invalidation, as @nicoburns mentioned above.

Loirooriol:
Should this be closed like https://github.com/servo/servo/pull/33543 ?

nicoburns:
I think this should be scoped down. I believe `:is()` and `:where()` (and possibly nested selectors?) should "just work" if enabled here, whereas `:has()`, `:nth-child()` require complex integration work.

mattsains:
@nicoburns , is there a tracking issue for the integration work required for the `:has()`, `:nth-child()`, and `:nth-last-child()` pseudoclasses? I'm not quite sure of what's necessary to get this supported.

I'm also wondering whether this issue is the same or distinct from the work required to support the `n of selector` syntax in `nth-child`.

Apologies if I've asked anything ignorant here, I'm not super familiar with this project yet

PR #82, opened 2024-10-14
Tracking PR: v0.1.0 release branch

nicoburns:
This PR isn't suitable for merging as-is, as it uses different package names to servo in some cases. But it documents the code used to make the [v0.1.0 release](https://crates.io/crates/stylo/0.1.0) and gives a good idea of the changes that will be required

Non-metadata changes required:

- Don't use cross-crate `include!()` macro for counter style atoms.
  - Fix in this PR: manually duplicate the file into both crates
  - Suggested fix: Update the Python script that generates the file to duplicate the file into both crates.
- Crates depending on git version of `webrender_api`. This is solely to import the `DevicePixel` marker type.
  - Fix in this PR: replace `DevicePixel` with a custom zero-variant enum (this is also what Gecko does).
  - Suggested fix 1: Same as in this PR + update Servo to deal with these types not being the same
  - Suggested fix 2: Release a new crates.io version of webrender_api and commit to do doing this regularly.
  - Suggested fix 3: Move `DevicePixel` to a new crate that can be shared.


to_shmem/Cargo.toml
```diff
@@ -1,7 +1,9 @@
 [package]
-name = "to_shmem"
+name = "stylo_to_shmem"
```
mrobinson, line 2: Why are you renaming `to_shmem` and `to_shmem_derive` here?

style_traits/lib.rs
```diff
@@ -33,11 +33,7 @@ extern crate to_shmem;
 #[macro_use]
 extern crate to_shmem_derive;
 #[cfg(feature = "servo")]
-extern crate webrender_api;
-#[cfg(feature = "servo")]
```
mrobinson, line 37: Doesn't removing the webrender dependency here break the Servo compilation?

style_traits/Cargo.toml
```diff
@@ -1,17 +1,20 @@
 [package]
-name = "style_traits"
-version = "0.0.1"
+name = "stylo_traits"
```
mrobinson, line 2: I think we shouldn't rename crates downstream and instead this should be managed in coordination with upstream.

style_static_prefs/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
-name = "static_prefs"
+name = "stylo_static_prefs"
 version = "0.1.0"
-edition = "2021"
 authors = ["The Servo Project Developers"]
+documentation = "https://docs.rs/stylo_static_prefs/"
+description = "Static configuration crate for Stylo CSS engine"
+repository = "https://github.com/servo/stylo"
+keywords = ["css", "style"]
```
mrobinson, line 8: These kind of changes belong upstream I think and then we can take them once we integrate them here.

servo_arc/Cargo.toml
```diff
@@ -1,14 +1,15 @@
 [package]
-name = "servo_arc"
+name = "stylo_arc"
```
mrobinson, line 2: It should be fine to continue using the `servo_arc` name for now.

atoms/counter_styles.rs
```diff
@@ -0,0 +1,61 @@
...
+    "decimal",
+    "decimal-leading-zero",
+    "arabic-indic",
+    "armenian",
+    "upper-armenian",
+    "lower-armenian",
+    "bengali",
+    "cambodian",
+    "khmer",
+    "cjk-decimal",
```
mrobinson, line 15: What's the purpose of this change?

mrobinson (commented):
This will need to be reworked to be against the `main` branch. It was auto-closed because it was written against the `upgrade` branch which gets deleted when an upgrade is finished.

atoms/counter_styles.rs
mrobinson: Ah, I see the comment in the PR description now. Sorry for the noise. I think this fix should go upstream probably.

to_shmem/Cargo.toml
nicoburns: This PR represents exactly what was published for the 0.1.0 release, for which I had to rename any crate that had already been published (as I didn't have publish access to publish a new version). It will need to look a little different for actually upstreaming. I just thought it would be good to have a reference of what was actually published (probably we should turn this into a branch in this repo).

style_traits/lib.rs
nicoburns: Yes it does. We need to come to some kind of consensus on what the best solution to this is.

My thought is that it probably does make sense to decouple the two libraries by removing the Webrender dependency here and fixup Servo to deal with the types being different. The advantage of doing this is that:

- Servo would be able to update Stylo and Webrender versions independently of each other
- Crates.io releases of Stylo and Webrender could follow independent release schedules.

IIRC there were 9 errors when I tried compiling Servo with this branch so it's not a huge task. I would be able to do this work if there's consensus that this an acceptable approach.

PR #86, opened 2024-10-29
Add support for `mix-blend-mode: plus-lighter`

mrobinson:


nicoburns (approved):
New version looks fine.

Might just be the diff, but it looks like the old version has a missing space in the gecko version?

Loirooriol:
> Might just be the diff, but it looks like the old version has a missing space in the gecko version?

Probably my mistake when disabling the keyword. It's not upstream.

PR #88, opened 2024-10-30
Add `stylo_taffy` crate

nicoburns:
This PR contains a new crate `stylo_taffy` which contains:

- Type conversions between `stylo` types and `taffy` types
- A function to convert  a `stylo::ComputeValues` into a `taffy::Style`
- A wrapper type `TaffyStyloStyle<T>` that wraps any `T: Deref<Target = ComputedValue>` and which implements Taffy's style traits, which allows it to be passed to Taffy's algorithms directly instead of Taffy's `Style` struct.

I believe this repository is the best place for this code to live as:

- It allows Servo, Blitz, and anybody else who wants to combine Stylo with Taffy to share this code
- It avoids any complications with having MIT/APACHE2.0 licensed code in the main Servo repo
- It means that any required fixups can happen directly as part of a Stylo sync, and there is not the additional step of having to update code in the Taffy repo first.

I can submit this upstream if there is agreement that this is the best place for this code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "style_derive",
     "style_static_prefs",
     "style_traits",
+    "stylo_taffy",
     "to_shmem",
     "to_shmem_derive",
 ]
--- a/stylo_taffy/Cargo.toml
+++ b/stylo_taffy/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "stylo_taffy"
+version = "0.1.0"
+license = "MIT OR APACHE 2.0"
+description = "Interop crate for the Stylo and Taffy crates"
+keywords = ["css", "layout"]
+categories = ["gui"]
+edition = "2021"
+repository = "https://github.com/servo/stylo"
+
+[features]
+default = ["std", "block", "flexbox", "grid"]
+std = ["taffy/std"]
+block = ["taffy/block_layout"]
+flexbox = ["taffy/flexbox"]
+grid = ["taffy/grid"]
+
+[dependencies]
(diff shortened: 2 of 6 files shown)


mrobinson:
I'm not sure this the best place to put this code. One of the big goals here is to reduce the diff with the upstream version of stylo.

nicoburns:
> I'm not sure this the best place to put this code. One of the big goals here is to reduce the diff with the upstream version of stylo.

Well I was imagining this would go upstream. That would of course be dependent on upstream accepting it.

Other options include:

- This crate exists in this form in the `servo/servo` repo. That would be mostly fine (although it would be kinda annoying to deal with having to run Servo's entire CI every time we want to make a change).
- This crate exists in this form in the Taffy repo. This would allow for code sharing, but would make Stylo syncs annoying, with an additional repo needing to updated.
- We include copies of this code in each repo it is used in, and copy-paste sync changes. This wouldn't be a huge deal from a code POV, but would either require Servo to include an MIT/APACHE2.0 files or Blitz to include an MPL ones (we could also triple-license but that would still require clear communication of the license which applies to the files).
- We include copies of this code in each repo with separate licenses, and accept that such code will diverge and that changes can't be copied between them.

mrobinson:
> This crate exists in this form in the servo/servo repo. That would be mostly fine (although it would be kinda annoying to deal with having to run Servo's entire CI every time we want to make a change).

Let's just do this. This is where the conversion is done, so it makes sense that the code lives there so that dead code analysis works properly. Note that we have this kind of conversion code other places in Servo as well.

nicoburns:
Ok, this code is now included as part of https://github.com/servo/servo/pull/32619

PR #91, opened 2024-11-21
Remove Servo-specific code from `malloc_size_of`

mrobinson:
Servo will use a forked version of `malloc_size_of` so that it does not
push depenencies down into Stylo.

Corresponding upstream change: https://phabricator.services.mozilla.com/D229764


jdm (approved):
Smart!

PR #92, opened 2024-11-23
Remove dependency on `webrender_api`

nicoburns:
I have submitted this upstream: https://bugzilla.mozilla.org/show_bug.cgi?id=1933010

We can actually land the Servo PR without landing this first. But we might want to land anyway this to ensure that Servo doesn't regress before the next stylo sync.

Servo PR: https://github.com/servo/servo/pull/34353

### DIFF
--- a/style_traits/Cargo.toml
+++ b/style_traits/Cargo.toml
@@ -11,7 +11,7 @@ name = "style_traits"
 path = "lib.rs"
 
 [features]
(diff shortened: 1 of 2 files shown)


Loirooriol (approved):

nicoburns:
This has been accepted (but not yet landed) upstream. But I'm going to land it here to so we can get the Servo PR merged (and make sure it doesn't regress before we next sync Stylo).

Loirooriol:
Ah, makes sense

PR #93, opened 2024-11-28
Enable `prefers-color-scheme` for Servo

nicoburns:
Tested with Blitz:

Uploading Screen Recording 2024-11-28 at 16.07.45.mov…


nicoburns:
I looked at upstreaming this change. However, this is a part of Stylo where we are currently maintaining quite a large diff, so my patch would not easily apply upstream (I could modify the implementation, but I think that would just cause conflicts when we sync). My proposal is that we land this here (along with the [corresponding servo PR](https://github.com/servo/servo/pull/34423)), and then I can look at upstreaming all of `style/servo` (or at least all of `servo/style/media_queries.rs`).

style/servo/media_queries.rs
```diff
@@ -357,8 +362,24 @@ fn eval_device_pixel_ratio(context: &Context) -> f32 {
     eval_resolution(context).dppx()
 }
 
+/// Values for the prefers-color-scheme media feature.
+#[derive(Clone, Copy, Debug, FromPrimitive, Parse, PartialEq, ToCss)]
+#[repr(u8)]
+#[allow(missing_docs)]
+pub enum PrefersColorScheme {
```
Loirooriol, line 369: Emilio prefers sharing these things between Servo and Gecko: https://phabricator.services.mozilla.com/D226265?vs=on&id=932339#inline-1256740

nicoburns: I actually have code for that already! (which I didn't push because I thought it might be better to do "one thing at a time"). I've pushed it now. Does that seem better?

Loirooriol (approved):

style/servo/media_queries.rs
Loirooriol: Yeah, well, no big deal either way if you prefer to leave that for a follow-up

PR #97, opened 2024-12-05
Eliminate thread pool mutex

nicoburns:
This was quite straightforward (just remove the Mutex and the calls to the lock methods). And `StyleThreadPool` already internally contains an `RwLock`. Perhaps this was previously required but isn't anymore?

Reduces the diff with upstream which doesn't have this Mutex.

Servo PR:
- https://github.com/servo/servo/pull/34480

### DIFF
--- a/style/global_style_data.rs
+++ b/style/global_style_data.rs
@@ -101,7 +101,7 @@ impl StyleThreadPool {
         }
         {
             // Drop the pool.
-            let _ = STYLE_THREAD_POOL.lock().unwrap().style_thread_pool.write().take();
+            let _ = STYLE_THREAD_POOL.style_thread_pool.write().take();
         }
 
         // Join spawned threads until all of the threads have been joined. This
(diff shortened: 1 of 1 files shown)


jdm (approved):

nicoburns:
It seems this [doesn't work](https://github.com/servo/servo/pull/34480#issuecomment-2518876863). Perhaps we could:
- Pop a comment into the source code linking to this analysis
- Move the Mutex into Servo `layout_thread` crates so that we don't need to maintain the diff here.

?

jdm:
If it's possible to move it into Servo's layout thread code, that seems like a worthwhile change.

nicoburns:
> If it's possible to move it into Servo's layout thread code, that seems like a worthwhile change.

@jdm Implemented over in https://github.com/servo/servo/pull/34480. I'll run the test suite again to check it works.

PR #99, opened 2024-12-10
Duplicate predefined counter styles (allows stylo_atoms crate to be published)

nicoburns:
This fix unblocks the `atoms` crate from being published on crates.io. I did look at having the python script which generates the file generate it into both locations. But that doesn't work because the `atoms` crate doesn't exist upstream. I think a simple duplication is probably the best approach. The rest of the static atom values are maintained manually anyway.


atoms/build.rs
```diff
@@ -22,7 +22,7 @@ fn main() {
             }
         }
     }
-    include!("../style/counter_style/predefined.rs");
+    include!("./predefined_counter_styles.rs");
```
mrobinson, line 25: A better way to do this would probably be to have the import script duplicate this file into the atoms directory.

nicoburns: I did think about that but this directory doesn't exist at import time either (at least not when calling git filter-repo) because it's not upstream. I suppose we could add a post rebase step to sync and squash this file? But that seems like it might be quite fragile?

Loirooriol (approved):

PR #102, opened 2024-12-26
Use truncation instead of rounding to resolve percentages into app units

Asun0204:
change conversion to appunit from float of percentage calcuation, prevent child box size is calculated too large.

Servo PR: https://github.com/servo/servo/pull/34714

### DIFF
--- a/style/values/computed/length_percentage.rs
+++ b/style/values/computed/length_percentage.rs
@@ -518,7 +518,11 @@ impl LengthPercentage {
     /// Returns the used value.
     #[inline]
     pub fn to_used_value(&self, containing_length: Au) -> Au {
-        Au::from(self.to_pixel_length(containing_length))
+        let length = self.to_pixel_length(containing_length);
+        if let Unpacked::Percentage(_) = self.unpack() {
+            return Au::from_f32_px_trunc(length.px());
+        }
+        Au::from(length)
     }
 
     /// Returns the used value as CSSPixelLength.
@@ -531,7 +535,11 @@ impl LengthPercentage {
     #[inline]
     pub fn maybe_to_used_value(&self, container_len: Option<Au>) -> Option<Au> {
         self.maybe_percentage_relative_to(container_len.map(Length::from))
-            .map(Au::from)
+            .map(if let Unpacked::Percentage(_) = self.unpack() {
+                |length: Length| Au::from_f32_px_trunc(length.px())
+            } else {
+                Au::from
+            })
     }
 
     /// If there are special rules for computing percentages in a value (e.g.


Loirooriol:
Since we want to test this before publishing a new release of app units, let's inline the code here for now.

Loirooriol:
BTW, when creating a PR it's better not to use your main branch.

Loirooriol (commented):
It may make more sense to modify this instead: https://github.com/servo/stylo/blob/7e529e8cf99502c99e431515de095d18548219d5/style/values/computed/length.rs#L496-L501

I thought that would be more likely to affect Firefox, but I have tried commenting that (and its callers) out, and Firefox compiles fine, so it's not used.

Asun0204:
I'm sorry about the code you commit is overwitten by mistake, can you push the code again? @Loirooriol

Loirooriol:
What code, https://github.com/servo/stylo/commit/83dad258d1a64a88a30468751774dec2169ebe67 ?

Loirooriol:
Sure, you can try that. But currently the logic first resolves LengthPercentage into a Length and then to an Au. You will need to remove the middle step.

Asun0204:
> Sure, you can try that. But currently the logic first resolves LengthPercentage into a Length and then to an Au. You will need to remove the middle step.

I don't known how to remove the middle step. Do I need to copy the percenrage calculation in `maybe_to_used_value`? And it will skip normallized operation provided by Length.

style/values/computed/length_percentage.rs
```diff
@@ -562,8 +562,11 @@ impl LengthPercentage {
     /// Convert the computed value into used value.
     #[inline]
     pub fn maybe_to_used_value(&self, container_len: Option<Au>) -> Option<Au> {
-        self.maybe_percentage_relative_to(container_len.map(Length::from))
-            .map(Au::from)
+        if let Unpacked::Percentage(_) = self.unpack() {
```
Loirooriol, line 565: Probably `Unpacked::Calc` should truncate too.

Loirooriol: Rounding for `calc()` is better according to https://phabricator.services.mozilla.com/D202746

Asun0204: The workflow that I just change the percenrage method using truncate has been passed. So should we merge app_units pr? Then the stylo pr could be merged. And do we need to modify the version?
https://github.com/Asun0204/servo/actions/runs/13385320860

Asun0204:
@Loirooriol The code is completed according to the solution discussed. Can you review the code again?

https://github.com/servo/stylo/pull/102
https://github.com/servo/app_units/pull/60

Loirooriol:
Update the Servo PR as requested: https://github.com/servo/servo/pull/34714#pullrequestreview-2641154294

Loirooriol (changes requested):
Don't revert e2c385d

Loirooriol (changes requested):
The PR is still including unrelated commits.

Loirooriol:
Remove https://github.com/servo/stylo/pull/102/commits/f588e855d77ba5ebc6ced99247b3163427f8c8b6

Loirooriol:
You need to rebase your changes on top of the latest `main` (0190fff).

Asun0204:
I completed it. There is only one commit base on the latest main.

style/values/computed/length_percentage.rs
```diff
@@ -530,8 +530,11 @@ impl LengthPercentage {
...
+            return self.maybe_percentage_relative_to(container_len.map(Length::from))
+                .map(|length| Au::from_f32_px_trunc(length.px()));
```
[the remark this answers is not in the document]
Asun0204: I push a commit to import app_units pr for compiling it.
It's ok now.

style/values/computed/length_percentage.rs
```diff
@@ -530,8 +530,11 @@ impl LengthPercentage {
...
+        }
+        self.maybe_percentage_relative_to(container_len.map(Length::from)).map(Au::from)
```
Loirooriol, line 537: No need to repeat `self.maybe_percentage_relative_to(container_len.map(Length::from))`

```suggestion
        self.maybe_percentage_relative_to(container_len.map(Length::from))
            .map(if let Unpacked::Percentage(_) = self.unpack() {
                |length| Au::from_f32_px_trunc(length.px())
            } else {
                Length::from
            })
    }
```

Loirooriol (changes requested):
Also you should cover `to_used_value` for consistency.

style/values/computed/length_percentage.rs
Asun0204: fixed.

style/values/computed/length_percentage.rs
```diff
@@ -518,6 +518,9 @@ impl LengthPercentage {
     /// Returns the used value.
     #[inline]
     pub fn to_used_value(&self, containing_length: Au) -> Au {
+        if let Unpacked::Percentage(_) = self.unpack() {
+            return Au::from_f32_px_trunc(self.to_pixel_length(containing_length).px());
+        }
         Au::from(self.to_pixel_length(containing_length))
```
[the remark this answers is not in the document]
Asun0204: fixed

Loirooriol:
You keep using "calcuation" instead of "calculation".
But just avoid this word since it seems to refer to `calc()` and this is about plain percentages.
Also "float of percentage" is confusing.

PR #103, opened 2025-01-06
Implement `::backdrop` pseudo-element parsing

nicoburns:
## Context

- Fixes https://github.com/servo/servo/issues/34820

## Notes

Github.com sets it's root-level CSS variables using a selector that includes a `::backdrop` pseudo selector. It's not quite this, but it's morally equivalent to `html, html ::backdrop`. And the presence of `::backdrop` causes the entire selector to fail to parse meaning that the variables never apply and styling is way off.

Changing the selector to `html, html ::non-existent-selector` causes the styles to fail to apply in Chrome too. So I think the strict parsing is correct per the spec. So I believe the only way to make this work is to implement parsing for `::backdrop`. I don't think we support creating a backdrop in a current Servo, but if we did support then we probably would support matching styles against it, so enabling the selector seems reasonable to me.

## Feedback wanted

I wasn't sure if it should be a `Lazy` pseudo-element. But `Lazy` is documented as being for regular but rarely used pseudo-elements which seemed to fit, whereas `PreComputed` is documented as being only for internal pseudo-elements which didn't seem to fit.

### DIFF
--- a/style/servo/selector_parser.rs
+++ b/style/servo/selector_parser.rs
@@ -50,6 +50,8 @@ pub enum PseudoElement {
     // them.  Also, make sure the UA sheet has the !important rules some of the
     // APPLIES_TO_PLACEHOLDER properties expect!
 
+    Backdrop,
+
     // Non-eager pseudos.
     DetailsSummary,
     DetailsContent,
@@ -82,6 +84,7 @@ impl ToCss for PseudoElement {
             After => "::after",
             Before => "::before",
             Selection => "::selection",
+            Backdrop => "::backdrop",
             DetailsSummary => "::-servo-details-summary",
             DetailsContent => "::-servo-details-content",
             ServoAnonymousBox => "::-servo-anonymous-box",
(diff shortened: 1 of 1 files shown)


Loirooriol (approved):
I think we don't support modal dialogs nor fullscreen, so it shouldn't be observable whether we support `::backdrop` or not, thus claiming we do isn't wrong :P

Not much sure about the pseudo cascade type either, but lazy sounds reasonable.

jdm:
> I think we don't support modal dialogs nor fullscreen, so it shouldn't be observable whether we support `::backdrop` or not, thus claiming we do isn't wrong :P
> 
> Not much sure about the pseudo cascade type either, but lazy sounds reasonable.

We do support the DOM full screen API, unless you're referring to something else.

nicoburns:
Does the fullscreen API use `::backdrop`? I thought it was just "top layer". I wouldn't expect fullscreen to have a backdrop because, well, it takes up the whole screen so you won't be be able to see anything behind.

Loirooriol:
Yes, the fullscreen API places elements in the top layer, in fact `::backdrop` was initially defined in the fullscreen spec: https://fullscreen.spec.whatwg.org/#css-pe-backdrop

PR #105, opened 2025-01-07
Parse :host selector

simonwuelker:
Stylo supports this selector, it just needs to be turned on.


jdm (approved):

simonwuelker:
I don't have merge permissions for stylo, so you need to push the button.

PR #109, opened 2025-01-13
Let animations & transitions fall back to discrete interpolation

Loirooriol:
When we are trying to interpolate two values that can't be interpolated according to the animation type of the CSS property, CSS animations should always fall back to discrete interpolation, and CSS transitions should also do so when `transition-behavior` is `allow-discrete`.

Servo PR: https://github.com/servo/servo/pull/34981

### DIFF
--- a/style/animation.rs
+++ b/style/animation.rs
@@ -95,11 +95,21 @@ impl PropertyAnimation {
     }
 
     /// Update the given animation at a given point of progress.
-    fn calculate_value(&self, progress: f64) -> Result<AnimationValue, ()> {
+    fn calculate_value(&self, progress: f64, allow_discrete: bool) -> Result<AnimationValue, ()> {
+        let progress = self.timing_function_output(progress);
         let procedure = Procedure::Interpolate {
-            progress: self.timing_function_output(progress),
+            progress,
         };
-        self.from.animate(&self.to, procedure)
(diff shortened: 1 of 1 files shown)


mrobinson:
Out of curiosity, why must we change platform-independent code for this? What is Gecko doing to support this?

Loirooriol:
Gecko seems to have its own thing to handle animations and transitions, e.g. in https://github.com/servo/servo/pull/30565/ I had to modify Stylo to stop discrete properties from transitioning, but Gecko already had the correct behavior.

But yeah this needs more investigation.

mrobinson:
@Loirooriol So does that mean that this code is unused by Gecko? If so this is fine to me, though we should start upstreaming our changes here if that's the case.

Loirooriol:
Yes, Emilio said that Firefox is compiling this code to avoid breaking it, but they are using their own thing: https://searchfox.org/mozilla-central/rev/a965e3c683ecc035dee1de72bd33a8d91b1203ed/servo/ports/geckolib/glue.rs#591-605

However, this PR isn't working correctly, transitions need to take `transition-behavior` into account.

mrobinson (approved):

Loirooriol:
Upstreaming changes in https://bugzilla.mozilla.org/show_bug.cgi?id=1945225

PR #110, opened 2025-01-14
Enable `quotes` property

xiaochengh:
For https://github.com/servo/servo/issues/34446

Actual behavioral changes will be implemented in another PR to the servo repository

### DIFF
--- a/style/properties/longhands/list.mako.rs
+++ b/style/properties/longhands/list.mako.rs
@@ -69,7 +69,6 @@ ${helpers.predefined_type(
     "Quotes",
     "computed::Quotes::get_initial_value()",
     engines="gecko servo",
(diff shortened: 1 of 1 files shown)


Loirooriol (approved):

mrobinson:
@xiaochengh Please don't use merge commits. Instead rebase and squash your work into a single commit.

xiaochengh:
Yeah I just noticed that the commits are totally messed up... And it's fixed.

Loirooriol:
> Yeah I just noticed that the commits are totally messed up

Yes that's because I force-pushed the main branch during an upgrade, this is the approach that we use but it has some downsides like messing with existing PRs.

PR #111, opened 2025-01-17
Enable overflow:clip for Servo

longvatrong111:
This PR help `overflow : clip` implementation in Servo
- Enable `overflow : clip` for Servo
- Implement `overflow-clip-margin`

Test case:
- /css/css-overflow/parsing/overflow-clip-margin.html
- /css/css-overflow/parsing/overflow-clip-margin-computed.html

Servo PR: https://github.com/servo/servo/pull/35103

cc: @xiaochengh

### DIFF
--- a/style/properties/longhands/margin.mako.rs
+++ b/style/properties/longhands/margin.mako.rs
@@ -30,12 +30,12 @@
 
 ${helpers.predefined_type(
     "overflow-clip-margin",
-    "Length",
-    "computed::Length::zero()",
-    parse_method="parse_non_negative",
-    engines="gecko",
+    "OverflowClipMargin",
+    "generics::box_::OverflowClipMargin::None",
+    engines="gecko servo",
     spec="https://drafts.csswg.org/css-overflow/#propdef-overflow-clip-margin",
     affects="overflow",
+    animation_type="discrete",
 )}
 
 % for index, side in enumerate(ALL_SIDES):
--- a/style/values/specified/box.rs
+++ b/style/values/specified/box.rs
@@ -10,14 +10,20 @@ use crate::values::generics::box_::{
     GenericContainIntrinsicSize, GenericLineClamp, GenericPerspective, GenericVerticalAlign,
     VerticalAlignKeyword,
 };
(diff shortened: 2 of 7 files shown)


style/values/specified/box.rs
```diff
@@ -936,6 +935,21 @@ pub enum OverflowAnchor {
 pub enum OverflowClipBox {
     PaddingBox,
     ContentBox,
+    BorderBox,
+}
+
+impl Parse for OverflowClipBox {
```
xiaochengh, line 941: I assume such trivial keyword parsers are auto-generated and we don't need a manual implementation here. Could you verify?

longvatrong111: Yes, I wanted to make it explicit. 
But I've just realized this is not necessary for `overflow : clip`. Should implement parse for overflow-clip-margin instead

style/properties/longhands/box.mako.rs
```diff
@@ -134,7 +134,7 @@ ${helpers.predefined_type(
 
 ${helpers.single_keyword(
     "-servo-overflow-clip-box",
-    "padding-box content-box",
+    "padding-box content-box border-box",
```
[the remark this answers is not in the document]
longvatrong111: I see. And it's not necessary for overflow:clip too. Will remove it.

style/properties/longhands/margin.mako.rs
```diff
@@ -30,12 +30,12 @@
 
 ${helpers.predefined_type(
     "overflow-clip-margin",
-    "Length",
-    "computed::Length::zero()",
-    parse_method="parse_non_negative",
-    engines="gecko",
+    "OverflowClipMargin",
```
Loirooriol, line 33: I would keep this as a length for now.
If afterwards you want to generalize it, it would be better to do it upstream in Firefox, and then I will sync it into here.
We are trying to minimize the difference with upstream, because otherwise syncing is very tedious.

longvatrong111: Can you check https://github.com/servo/stylo/pull/114?
I'll comeback with `visual-box` for `overflow-clip-margin` later.

PR #113, opened 2025-01-24
add cancel to static_atoms

keithamus:
Needed for dialog events in Servo.


lukewarlow:
For the record this is needed for the 'cancel' event which is used by the dialog element along with close watchers.

jdm (approved):

PR #115, opened 2025-01-30
Avoid starting transitions if values can't be interpolated

Loirooriol:
A property may have an animation type different than 'discrete', but still not be able to interpolate some values. In that case animations fall back to discrete interpolation, but transitions can only do so if allowed by `transition-behavior`.

We were already handling that well, but we were still triggering `transitionstart` events even if no transition ended up happening. This patchs prevents the creation of the transition, and simplifies a bunch of code since we no longer need to check whether the transition is allowed while the transition runs.


Loirooriol:
BTW, the fix matches Gecko's logic: https://searchfox.org/mozilla-central/rev/548b6981501f59e3c9f2f7851c013e7d53c4e72f/servo/ports/geckolib/glue.rs#1262

mrobinson (approved):
Yeah, this is much cleaner. Thanks!

PR #116, opened 2025-01-31
Let `resolve_color()` take its parameter by reference

Loirooriol:
This way the callers don't have to clone it if they don't have ownership or want to use the value later.

Servo PR: https://github.com/servo/servo/pull/35247

### DIFF
--- a/style/properties/properties.mako.rs
+++ b/style/properties/properties.mako.rs
@@ -1874,9 +1874,9 @@ impl ComputedValues {
     ///
     /// Usage example:
     /// let top_color =
-    ///   style.resolve_color(style.get_border().clone_border_top_color());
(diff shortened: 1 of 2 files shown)


nicoburns (approved):

style/properties/properties.mako.rs
```diff
@@ -1874,9 +1874,9 @@ impl ComputedValues {
     ///
     /// Usage example:
     /// let top_color =
-    ///   style.resolve_color(style.get_border().clone_border_top_color());
+    ///   style.resolve_color(&style.get_border().clone_border_top_color());
```
Loirooriol, line 1877: For Servo this would be a better example:
```rs
style.resolve_color(&style.get_border().border_top_color);
```
but that doesn't work in Gecko so I guess it's better to leave it as is.

PR #117, opened 2025-02-03
Update derive_more to v2.0.0

JelteF:
This updates `derive_more` to the most recent version. The main reason
I'm interested in doing this is so that the latest version will become
available on the Rust Playground. The `derive_more` crate is being
pulled due to selectors, but it pulls in an outdated version due to
selectors depending on that version.

See additional information here: https://github.com/rust-lang/rust-playground/pull/1133


Loirooriol:
In general we try to not diverge from upstream if possible.
Could you do the upgrade in Gecko? Then we will import the change here.
See https://hg.mozilla.org/mozilla-central/rev/6f3f7162c687dcdc926008cb62f320f10735d917 for reference

JelteF:
Thanks for the pointer. I submitted it here now: https://phabricator.services.mozilla.com/D236782

mrobinson:
Closing this one as review is now happening upstream. Thanks @JelteF!

Loirooriol:
The last release includes your patch from upstream

Loirooriol:
FYI, it's going to be removed: https://phabricator.services.mozilla.com/D302241

nicoburns:
> FYI, it's going to be removed: https://phabricator.services.mozilla.com/D302241

It's only be removed from `selectors`, not from the `stylo`/`style` crate (selectors only uses it in one place, and can otherwise be made to work without proc macros, whereas `style` uses it all over the place and has other proc macro deps)

Loirooriol:
Yes, but it's included in the rust playground via selectors. Anyways, trpl isn't planning to upgrade scraper, so doesn't really matter.

PR #121, opened 2025-02-07
backdrop-filter: should create stacking context in servo.

JoshuaBrest:
This needs to create a stacking context to fix some websites that don't specify `isolation` in addition to `backdrop-filter`.


style/properties/longhands/effects.mako.rs
```diff
@@ -59,7 +59,8 @@ ${helpers.predefined_type(
     "backdrop-filter",
     "Filter",
     None,
-    engines="gecko",
+    engines="gecko servo",
+    servo_pref="layout.backdrop-filter.creates-stacking-context",
```
nicoburns, line 63: This should probably just be `layout_backdrop_filter` or `layout.backdrop_filter`

nicoburns (changes requested):

style/properties/longhands/effects.mako.rs
mrobinson: This probably doesn't need a preference at all.

JoshuaBrest: @mrobinson resolved!!

style/properties/longhands/effects.mako.rs
```diff
@@ -59,7 +59,7 @@ ${helpers.predefined_type(
     "backdrop-filter",
     "Filter",
     None,
-    engines="gecko",
+    engines="gecko servo",
```
Loirooriol, line 62: I do think this should have a runtime flag `layout.backdrop-filter.enabled` if you don't plan to implement it beyond creating a stacking context.

Also the title should be more like "Enable `backdrop-filter` on Servo, behind runtime flag". The stacking context should be mentioned in the Servo PR instead.

Loirooriol (changes requested):

style/properties/longhands/effects.mako.rs
JoshuaBrest: > I do think this should have a runtime flag `layout.backdrop-filter.enabled` if you don't plan to implement it beyond creating a stacking context.
> 
> Also the title should be more like "Enable `backdrop-filter` on Servo, behind runtime flag". The stacking context should be mentioned in the Servo PR instead.

Can one person just tell me which one is wanted? Because I had another one, and they told me I should not add a runtime flag.

mrobinson: The thing is, if the goal is to fix web content (which is using this feature to ensure a stacking context), the fix will only be effective if the there is no preference or the preference is turned on. The idea with adding the preference is that it is turned off by default, which means that for most emedders this will fix absolutely nothing.

We should either:

 1.  add this feature completely within a reasonably short amount of time, turning the preference on when it's complete. 
 2. just don't expose a preference so that a stacking context is created unconditionally with a plan to actually add support for the feature later.

Adding a preference where only some embedders or WPT test runs have a special, slightly more correct behavior, seems like the worst choice here.

mrobinson: > The thing is, if the goal is to fix web content (which is using this feature to ensure a stacking context), the fix will only be effective if the there is no preference or the preference is turned on. The idea with adding the preference is that it is turned off by default, which means that for most emedders this will fix absolutely nothing.
> 
> We should either:
> 
>     1. add this feature completely within a reasonably short amount of time, removing the preference  when it's complete.
> 
>     2. just don't expose a preference so that a stacking context is created unconditionally with a plan to actually add support for the feature later.
> 
> 
> Adding a preference where only some embedders or WPT test runs have a special, slightly more correct behavior, seems like the worst choice here.

Loirooriol: So not sure if I understood the reasoning for this. Is it because you thought that it would be a good first bug to fix, or is it for web compat? You mentioned you have used `backdrop-filter: blur(0px)` to establish a stacking context, but has that been in like personal projects, or is this affecting actual websites? Are you aware of it being a common pattern among other developers?

If you are just looking for a good first bug about establishing stacking contexts, it may be better to implement `isolation: isolate` (https://github.com/servo/servo/issues/35337). If you are trying to fix web content then it can make sense to do this without a pref.

JoshuaBrest: > So not sure if I understood the reasoning for this. Is it because you thought that it would be a good first bug to fix, or is it for web compat? You mentioned you have used backdrop-filter: blur(0px) to establish a stacking context, but has that been in like personal projects, or is this affecting actual websites? Are you aware of it being a common pattern among other developers?

I feel like more people use this than the isolate. @Loirooriol

Loirooriol: I only mentioned `isolation: isolate` in case that you are not trying to fix web content.

I checked WebKit, it seems that they made `backdrop-filter` establish a stacking context when they implemented the property, not before or after: https://github.com/WebKit/Webkit/commit/4b6363d3d2519ca9f72750e42b06a4e9f962adbf#diff-7cc093e2dd898772c51e0313408a4e9441d4d62049786b0dcac3089a226ef995R125

```diff
     // Returns true if this renderer requires a new stacking context.
-    bool createsGroup() const { return isTransparent() || hasMask() || hasFilter() || hasBlendMode(); }
+    bool createsGroup() const { return isTransparent() || hasMask() || hasFilter() || hasBackdropFilter() || hasBlendMode(); }
```

So was the problem with the broken sites a lack of stacking context, or a lack of `backdrop-filter`?

JoshuaBrest:
https://github.com/servo/servo/pull/35415

style/properties/longhands/effects.mako.rs
JoshuaBrest: > I only mentioned `isolation: isolate` in case that you are not trying to fix web content.
> 
> I checked WebKit, it seems that they made `backdrop-filter` establish a stacking context when they implemented the property, not before or after: [WebKit/WebKit@4b6363d#diff-7cc093e2dd898772c51e0313408a4e9441d4d62049786b0dcac3089a226ef995R125](https://github.com/WebKit/WebKit/commit/4b6363d3d2519ca9f72750e42b06a4e9f962adbf#diff-7cc093e2dd898772c51e0313408a4e9441d4d62049786b0dcac3089a226ef995R125)
> 
> ```diff
>      // Returns true if this renderer requires a new stacking context.
> -    bool createsGroup() const { return isTransparent() || hasMask() || hasFilter() || hasBlendMode(); }
> +    bool createsGroup() const { return isTransparent() || hasMask() || hasFilter() || hasBackdropFilter() || hasBlendMode(); }
> ```
> 
> So was the problem with the broken sites a lack of stacking context, or a lack of `backdrop-filter`?

https://bugs.webkit.org/show_bug.cgi?id=262926

Loirooriol: > https://bugs.webkit.org/show_bug.cgi?id=262926

That testcase is about establish a containing block, not a stacking context. Still relevant if you implement both things in Servo, but your argument would be more compelling without mixing up concepts.

> an organization member said that I should not be adding the runtime flag

I'm also an organization member BTW. But fine, if you say some websites were affected and Martin is fine I won't oppose.

But you still need to fix the commit message. This is just enabling the property, not establishing a stacking context, leave that for Servo.

mrobinson:
Support for `backdrop-filter` for Servo was enabled upstream, so this isn't needed any longer.

Loirooriol:
That was done in #128, but it's still disabled at runtime:

https://github.com/servo/stylo/blob/03ffc85c45f53a3b11994f7c14adce7b90aa9c34/style/properties/longhands/effects.mako.rs#L69

PR #122, opened 2025-02-13
[feature](servo-text-overflow): Enable text-overflow for layout-2020 in servo

d-desiatkin:
This PR should be merged together with servo PR:
https://github.com/servo/servo/pull/35444

In case of compiling on windows, check following PR:
https://github.com/servo/stylo/pull/123

PR #123, opened 2025-02-13
[bugfix](stylo): Fix "bare CR not allowed in doc-comment" windows

d-desiatkin:
In case of building servo on Windows platform; Specifically after the following PR: https://github.com/servo/stylo/pull/122
I keep getting following error message:  "bare CR not allowed in doc-comment".
Changes bellow helped me to resolve the problem


Loirooriol:
So apparently, without `newline=''`, `f.write` will convert newlines to the default line separator on that OS?

I guess Firefox is affected too? May be better to fix it upstream and then it will be imported here.

PR #128, opened 2025-03-01
Enable will-change property for servo

yezhizhen:
Needed for https://github.com/servo/servo/issues/35470

Servo PR: https://github.com/servo/servo/pull/35787

cc @xiaochengh

### DIFF
--- a/style/values/specified/box.rs
+++ b/style/values/specified/box.rs
@@ -968,7 +968,7 @@ pub struct WillChange {
     /// A bitfield with the kind of change that the value will create, based
     /// on the above field.
     #[css(skip)]
-    bits: WillChangeBits,
+    pub bits: WillChangeBits,
 }
 
 impl WillChange {
@@ -1022,10 +1022,6 @@ bitflags! {
     }
 }
 
-#[cfg(feature="servo")]
-fn change_bits_for_longhand(_longhand: LonghandId) -> WillChangeBits { WillChangeBits::empty() }
-
-#[cfg(feature = "gecko")]
 fn change_bits_for_longhand(longhand: LonghandId) -> WillChangeBits {
     match longhand {
         LonghandId::Opacity => WillChangeBits::OPACITY,
--- a/style/properties/longhands/box.mako.rs
+++ b/style/properties/longhands/box.mako.rs
@@ -233,11 +233,12 @@ ${helpers.predefined_type(
     "offset-path",
     "OffsetPath",
     "computed::OffsetPath::none()",
-    engines="gecko",
+    engines="gecko servo",
     flags="CAN_ANIMATE_ON_COMPOSITOR",
     spec="https://drafts.fxtf.org/motion-1/#offset-path-property",
     servo_restyle_damage="reflow_out_of_flow",
     affects="overflow",
+    servo_pref="layout.unimplemented",
 )}
 
 // Motion Path Module Level 1
@@ -476,10 +477,11 @@ ${helpers.predefined_type(
     "contain",
     "Contain",
     "specified::Contain::empty()",
-    engines="gecko",
+    engines="gecko servo",
     animation_type="none",
     spec="https://drafts.csswg.org/css-contain/#contain-property",
     affects="layout",
+    servo_pref="layout.unimplemented",
 )}
 
 ${helpers.predefined_type(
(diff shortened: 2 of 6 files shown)


Loirooriol (approved):
LGTM, but this will need a Servo PR with the implementation before landing.

yezhizhen:
I added a function, `fn contains_critical_properties`. I believe it is needed, to check if certain values are contained in `will-change`

I am also referring to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context to check the list of properties whose non-initial value would create a stacking context.

style/values/specified/box.rs
```diff
@@ -977,6 +977,22 @@ impl WillChange {
     pub fn auto() -> Self {
         Self::default()
     }
+
+    #[inline]
+    /// Check if any property would create stacking context 
+    /// From <https://www.w3.org/TR/css-will-change/#valdef-will-change-custom-ident>:
+    /// > If any non-initial value of a property would create a stacking context on the element, 
+    /// > specifying that property in will-change must create a stacking context on the element.
+    pub fn contains_critical_properties(&self) -> bool {
```
Loirooriol, line 986: The logic needs to be different for establishing a stacking context vs containing block.

I would just do the check in layout.

yezhizhen: Then I guess need to change the visibility of fields in `WillChange` struct? (i.e., pub bits)

yezhizhen:
I am almost done. I have one question tho: `change_bits_for_longhand` for servo simply returns empty `WillChangeBits`. In that case, `STACKING_CONTEXT_UNCONDITIONAL` is never enabled in servo, even with filter/backdrop-filter.

https://github.com/servo/stylo/blob/8c57436148122c32b3cc3d40ee7c71806af1d087/style/values/specified/box.rs#L1023-L1034

I want to enable it for servo. But would instead get errors like:
```
error[E0599]: no variant or associated item named `Contain` found for enum `generated::LonghandId` in the current scope
     --> D:\stylo\style\values\specified\box.rs:1026:21
      |
1026  |         LonghandId::Contain => WillChangeBits::CONTAIN,
      |                     ^^^^^^^ variant or associated item not found in `LonghandId`
```

Do you think I should enable these? If so, how?

yezhizhen:
I rebased on latest `main` Stylo branch.

style/values/specified/box.rs
```diff
@@ -1045,6 +1041,11 @@ fn change_bits_for_longhand(longhand: LonghandId) -> WillChangeBits {
             WillChangeBits::STACKING_CONTEXT_UNCONDITIONAL | WillChangeBits::FIXPOS_CB_NON_SVG
         },
         LonghandId::ViewTransitionName => WillChangeBits::VIEW_TRANSITION_NAME,
+        #[cfg(feature = "servo")]
```
Loirooriol, line 1044: Trailing space

```suggestion
        #[cfg(feature = "servo")]
```

style/values/specified/box.rs
```diff
@@ -1045,6 +1041,11 @@ fn change_bits_for_longhand(longhand: LonghandId) -> WillChangeBits {
             WillChangeBits::STACKING_CONTEXT_UNCONDITIONAL | WillChangeBits::FIXPOS_CB_NON_SVG
         },
         LonghandId::ViewTransitionName => WillChangeBits::VIEW_TRANSITION_NAME,
+        #[cfg(feature = "servo")] 
+        LonghandId::MixBlendMode |
+        LonghandId::Isolation |
+        LonghandId::ClipPath => WillChangeBits::STACKING_CONTEXT_UNCONDITIONAL,
+        #[cfg(feature = "gecko")]
```
Loirooriol, line 1048: Trailing space

style/values/specified/box.rs
Loirooriol, line 1044: Why the inconsistency for `LonghandId::MaskImage`? Enable it like the others, or don't enable any of them.

Loirooriol (changes requested):

style/values/specified/box.rs
yezhizhen: On it. I was (and still is) confused how does following connects with MaskImage (which is also not found in style built). Guess `properties/build.py` is doing some magic.
```
${helpers.predefined_type(
    "mask-image",
    "Image",
    engines="gecko servo",
    initial_value="computed::Image::None",
    initial_specified_value="specified::Image::None",
    parse_method="parse_with_cors_anonymous",
    spec="https://drafts.fxtf.org/css-masking-1/#propdef-mask-image",
    vector=True,
    extra_prefixes="webkit",
    animation_type="discrete",
    affects="paint",
    servo_pref="layout.unimplemented",
)}
```

Loirooriol (approved):

PR #131, opened 2025-03-09
Update metadata and rename `dom`, `atoms`, `style_config` and `style_static_prefs` crates

nicoburns:
**This PR only affects crates that exist only in this repository**
(crates that are synced from upstream will be handled separately).

---

These changes rename crates and add `Cargo.toml` metadata in preparation for publishing to crates.io.

Servo PR: https://github.com/servo/servo/pull/35898

#### Renames
- `dom` is renamed to `stylo_dom`
- `style_static_prefs` renamed to `stylo_static_prefs`
- `style_config` renamed to `stylo_config`

#### Metadata updates
Cargo.toml metadata keys such as `version`, `documentation`, `description`, `repository`, and `keywords` have been added and/or updated for the following crates:
- `dom`/`stylo_dom`
- `style_static_prefs`/`stylo_static_prefs`
- `style_config`/`stylo_config`
- `servo_atoms`


atoms/Cargo.toml
```diff
@@ -1,10 +1,12 @@
 [package]
 name = "servo_atoms"
```
mrobinson, line 2: Maybe this should be called `stylo_atoms`?

atoms/Cargo.toml
```diff
@@ -1,10 +1,12 @@
...
-version = "0.0.1"
+version = "0.1.0"
```
mrobinson, line 3: What's the reasoning for the version bump here?

nicoburns: No particular reason other than it's usual for the first published version of a crates to be `0.1.0`. That allows for the possibility of non-breaking releases (I believe otherwise Cargo will treat the patch version as breaking).

atoms/Cargo.toml
nicoburns: Yeah, would be happy to go with that name.

atoms/Cargo.toml
```diff
@@ -1,10 +1,12 @@
 [package]
-name = "servo_atoms"
-version = "0.0.1"
+name = "stylo_atoms"
+version = "0.1.0"
 authors = ["The Servo Project Developers"]
+documentation = "https://docs.rs/stylo_atoms/"
+description = "Interned string type for the servo and stylo projects"
```
Loirooriol, line 6: Not a native speaker, but "internal" sounds less archaic?

```suggestion
description = "Internal string type for the servo and stylo projects"
```

style_static_prefs/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
-name = "static_prefs"
+name = "stylo_static_prefs"
 version = "0.1.0"
-edition = "2021"
 authors = ["The Servo Project Developers"]
+documentation = "https://docs.rs/stylo_static_prefs/"
+description = "Static configuration crate for Stylo CSS engine"
```
Loirooriol, line 6: It's weird to have he same description as `stylo_config`

dom/Cargo.toml
```diff
@@ -1,10 +1,13 @@
 [package]
-name = "dom"
-version = "0.0.1"
+name = "stylo_dom"
+version = "0.1.0"
 authors = ["The Servo Project Developers"]
+documentation = "https://docs.rs/stylo_dom/"
+description = "Support crate for Stylo CSS engine"
```
Loirooriol, line 6: Maybe mention the DOM, it's a bit weird to have such a generic description.

atoms/Cargo.toml
nicoburns: That would change the meaning here I think. The idea is to refer to the concept of [string interning](https://en.wikipedia.org/wiki/String_interning) rather than indicate that this library is internal to Servo.

dom/Cargo.toml
nicoburns: Yeah. I guess so. This crate barely even has anything to do with the DOM though. It's just the case that it contains types that are defined within Gecko's DOM module (these types previously lived within `style_traits` which makes more sense from a Servo perspective).

style_static_prefs/Cargo.toml
nicoburns: I have updated the `stylo_config` description as this is correct for this crate, but `stylo_config` is not "static"

dom/Cargo.toml
nicoburns: Updated to mention to the DOM.

nicoburns:
I think I've addressed all the review comments so far. I've also added a last commit that renames the directories to match the new crates names. I think that's worth doing (esp. as we aren't storing history for these crates anyway), but it does make the diff a bit harder to read so I can remove it if preferred.

mrobinson:
Looks good to me, but I'll wait for @Loirooriol to also give a review. Thanks!

Loirooriol (approved):

PR #134, opened 2025-03-10
Remove unnecessary imports

nicoburns:
1. Import is unnecessary as the same type is imported at the top of the file.
2. In general, you don't need to import Debug derive macro

These imports do not exist upstream so this helps to reduce the diff.

### DIFF
--- a/style/properties/properties.mako.rs
+++ b/style/properties/properties.mako.rs
@@ -1949,7 +1949,6 @@ impl ComputedValues {
 
     /// Get the initial computed values.
     pub fn initial_values_with_font_override(default_font: super::style_structs::Font) -> Arc<Self> {
-        use crate::logical_geometry::WritingMode;
         use crate::computed_value_flags::ComputedValueFlags;
         use servo_arc::Arc;
         use super::{ComputedValues, ComputedValuesInner, longhands, style_structs};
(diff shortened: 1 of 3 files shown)


style/values/specified/length.rs
```diff
@@ -1231,7 +1231,6 @@ impl NoCalcLength {
     /// absolute or (if a font metrics getter is provided) font-relative units.
     #[cfg(feature = "gecko")]
     #[inline]
-    #[cfg(feature = "gecko")]
```
nicoburns, line 1234: Note: duplicate attribute.

style/properties/properties.mako.rs
```diff
@@ -1949,7 +1949,6 @@ impl ComputedValues {
 
     /// Get the initial computed values.
     pub fn initial_values_with_font_override(default_font: super::style_structs::Font) -> Arc<Self> {
-        use crate::logical_geometry::WritingMode;
```
nicoburns, line 1952: Already imported at top of file

style/values/specified/box.rs
```diff
@@ -15,7 +15,7 @@ use crate::values::specified::{AllowQuirks, Integer, NonNegativeNumberOrPercenta
 use crate::values::CustomIdent;
 use cssparser::Parser;
 use num_traits::FromPrimitive;
-use std::fmt::{self, Debug, Write};
+use std::fmt::{self, Write};
```
nicoburns, line 18: It is not necessary to import `std::fmt::Debug` just to use the derive macro.

Loirooriol (approved):

mrobinson (approved):

PR #142, opened 2025-03-13
Sync upstream branch hourly rather than daily

nicoburns:
Syncing isn't a taxing activity for either our CI (typically completes in ~3 mins) or Mozilla (just does the equivalent of a single git pull). And it's useful to have our `upstream` up-to-date with the latest commits as soon as they are available.

So this changes our config to sync hourly rather than daily

### DIFF
--- a/.github/workflows/sync-upstream.yml
+++ b/.github/workflows/sync-upstream.yml
@@ -2,7 +2,7 @@ name: Sync upstream with mozilla-central
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


sagudev (approved):

Loirooriol (commented):
Every hours seems quite unnecessary, since AFAIK mozilla-central only lands patches from autoland every 12 hours or so?

PR #145, opened 2025-03-14
Rework `ElementAnimationSet::start_transition_if_applicable` to be more like the specification

yezhizhen:
This improves the behavior of transition toggling, cancellation, and
delay. In particular, there are many improvements with how transitions
are started when replacing an existing one.

Servo PR: https://github.com/servo/servo/pull/35978
    
Co-authored-by: Martin Robinson <mrobinson@igalia.com>
Signed-off-by: Euclid Ye <yezhizhenjiakang@gmail.com>

### DIFF
--- a/style/animation.rs
+++ b/style/animation.rs
@@ -25,7 +25,7 @@ use crate::style_resolver::StyleResolverForElement;
 use crate::stylesheets::keyframes_rule::{KeyframesAnimation, KeyframesStep, KeyframesStepValue};
 use crate::stylesheets::layer_rule::LayerOrder;
 use crate::values::animated::{Animate, Procedure};
-use crate::values::computed::{Time, TimingFunction};
+use crate::values::computed::TimingFunction;
 use crate::values::generics::easing::BeforeFlag;
 use crate::values::specified::TransitionBehavior;
 use crate::Atom;
@@ -55,31 +55,6 @@ impl PropertyAnimation {
     pub fn property_id(&self) -> PropertyDeclarationId {
         debug_assert_eq!(self.from.id(), self.to.id());
         self.from.id()
-    }
-
-    fn from_property_declaration(
-        property_declaration: &PropertyDeclarationId,
-        timing_function: TimingFunction,
-        duration: Time,
-        old_style: &ComputedValues,
-        new_style: &ComputedValues,
-    ) -> Option<PropertyAnimation> {
-        // FIXME(emilio): Handle the case where old_style and new_style's writing mode differ.
-        let property_declaration = property_declaration.to_physical(new_style.writing_mode);
-        let from = AnimationValue::from_computed_values(property_declaration, old_style)?;
-        let to = AnimationValue::from_computed_values(property_declaration, new_style)?;
-        let duration = duration.seconds() as f64;
-
-        if from == to || duration == 0.0 {
-            return None;
-        }
-
-        Some(PropertyAnimation {
-            from,
-            to,
-            timing_function,
-            duration,
-        })
     }
 
     /// The output of the timing function given the progress ration of this animation.
@@ -133,6 +108,11 @@ impl AnimationState {
     fn needs_to_be_ticked(&self) -> bool {
         *self == AnimationState::Running || *self == AnimationState::Pending
     }
+}
+
+enum IgnoreTransitions {
+    Canceled,
+    CanceledAndFinished,
 }
 
 /// This structure represents a keyframes animation current iteration state.
@@ -762,6 +742,31 @@ pub struct Transition {
 }
 
 impl Transition {
+    fn new(
+        start_time: f64,
+        delay: f64,
+        duration: f64,
+        from: AnimationValue,
+        to: AnimationValue,
+        timing_function: &TimingFunction,
+    ) -> Self {
+        let property_animation = PropertyAnimation {
+            from: from.clone(),
+            to,
+            timing_function: timing_function.clone(),
+            duration,
+        };
+        Self {
+            start_time,
+            delay,
+            property_animation,
+            state: AnimationState::Pending,
+            is_new: true,
+            reversing_adjusted_start_value: from,
+            reversing_shortening_factor: 1.0,
+        }
+    }
+
     fn update_for_possibly_reversed_transition(
         &mut self,
         replaced_transition: &Transition,
@@ -894,7 +899,7 @@ impl ElementAnimationSet {
             }
         }
 
-        if let Some(map) = self.get_value_map_for_active_transitions(now) {
+        if let Some(map) = self.get_value_map_for_transitions(now, IgnoreTransitions::Canceled) {
             for value in map.values() {
                 value.set_in_style_for_servo(mutable_style);
             }
@@ -1010,9 +1015,22 @@ impl ElementAnimationSet {
             self,
         );
 
-        // Cancel any non-finished transitions that have properties which no longer transition.
+        // Cancel any non-finished transitions that have properties which no
+        // longer transition.
+        //
+        // Step 3 in https://drafts.csswg.org/css-transitions/#starting:
+        // > If the element has a running transition or completed transition for
+        // > the property, and there is not a matching transition-property value,
+        // > then implementations must cancel the running transition or remove the
+        // > completed transition from the set of completed transitions.
+        //
+        // TODO: This is happening here as opposed to in
+        // `start_transition_if_applicable` as an optimization, but maybe this
+        // code should be reworked to be more like the specification.
         for transition in self.transitions.iter_mut() {
-            if transition.state == AnimationState::Finished {
+            if transition.state == AnimationState::Finished
+                || transition.state == AnimationState::Canceled
+            {
                 continue;
             }
             if transitioning_properties.contains(transition.property_animation.property_id()) {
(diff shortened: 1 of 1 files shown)


nicoburns (dismissed):
This needs to be rebased on top of the latest `main`

style/animation.rs
```diff
@@ -1015,11 +1015,55 @@ impl ElementAnimationSet {
...
+            // must cancel the running transition or remove the completed transition from the set of completed transitions."
+            if !transitioning_properties.contains(transition.property_animation.property_id()) {
+                transition.state = AnimationState::Canceled;
+                self.dirty = true;
                 continue;
             }
-            transition.state = AnimationState::Canceled;
-            self.dirty = true;
+            
+            let transition_property_id = transition.property_animation.property_id();
```
Loirooriol, line 1028: Move this upwards so that you can use `if !transitioning_properties.contains(transition_property_id)`

style/animation.rs
```diff
@@ -1015,11 +1015,55 @@ impl ElementAnimationSet {
...
+            // and the end value of the running transition is not equal to the value of the property in the after-change style, then:"
+            if transition.property_animation.to != after_change_to {
+                // index for after-change transition declaration
+                let index = after_change_style.transition_properties().position(|declared_transition| declared_transition.property
+                    .as_borrowed()
+                    .to_physical(after_change_style.writing_mode) == transition_property_id).unwrap();
+
+                let now = context.current_time_for_animations;
+                let after_change_style = after_change_style.get_ui();
+                let allow_discrete: bool = after_change_style.transition_behavior_mod(index) == TransitionBehavior::AllowDiscrete;
```
Loirooriol, line 1042:
```suggestion
                let allow_discrete = after_change_style.transition_behavior_mod(index) == TransitionBehavior::AllowDiscrete;
```

style/animation.rs
```diff
@@ -1015,11 +1015,55 @@ impl ElementAnimationSet {
...
+                let not_transitionable = !transition_property_id.is_animatable() || (!allow_discrete && transition_property_id.is_discrete_animatable());
```
Loirooriol, line 1043: A property like `width` is animatable and not discretely animatable. But still, some values are not transitionable, e.g. from `0px` to `max-content` with `transition-behavior: normal`

See https://github.com/servo/stylo/pull/115

yezhizhen:
Test re-running
https://github.com/yezhizhen/servo/actions/runs/13861265805

style/animation.rs
```diff
@@ -70,7 +70,7 @@ impl PropertyAnimation {
         let to = AnimationValue::from_computed_values(property_declaration, new_style)?;
         let duration = duration.seconds() as f64;
 
-        if from == to || duration == 0.0 {
+        if from == to || duration <= 0.0 {
```
yezhizhen, line 73: I noticed another bug: set `target.style.transition` to 0 duration, with 2s delay. Then the later change of property should occur instantenously, after 2 seconds delay.

But Servo has no delay in this case. I think this can be easily fixed by using combined duration.

style/animation.rs
```diff
@@ -1012,14 +1011,61 @@ impl ElementAnimationSet {
...
+                transition.state = AnimationState::Canceled;
+                self.dirty = true;
                 continue;
             }
-            transition.state = AnimationState::Canceled;
-            self.dirty = true;
+            
+            let after_change_to = AnimationValue::from_computed_values(transition_property_id, &after_change_style).unwrap();
+            // Step 4
+            // "If the element has a running transition for the property, there is a matching transition-property value,
```
yezhizhen, line 1030: After pondering, I believe I can merge step 4.1 and 4.2 into `fn start_transition_if_applicable` for better performance and clearity.

It is redundant and costly to recompute the index of declared transition property here.

style/animation.rs
```diff
@@ -1012,14 +1011,61 @@ impl ElementAnimationSet {
...
+            
+            let after_change_to = AnimationValue::from_computed_values(transition_property_id, &after_change_style).unwrap();
+            // Step 4
+            // "If the element has a running transition for the property, there is a matching transition-property value, 
+            // and the end value of the running transition is not equal to the value of the property in the after-change style, then:"
+            if transition.property_animation.to != after_change_to {
+                // index for after-change transition declaration
+                let index = after_change_style.transition_properties().position(|declared_transition| declared_transition.property
+                    .as_borrowed()
+                    .to_physical(after_change_style.writing_mode) == transition_property_id).unwrap();
```
yezhizhen, line 1036: After pondering, I believe I can merge step 4.1 and 4.2 into `fn start_transition_if_applicable` for better performance and clearity.

It is costly to compute the index of declared transition property here.

yezhizhen:
What should I rebase on now? I got many conflicts when rebasing on main.

Loirooriol:
I force-pushed the main branch to sync changes from upstream, which is not ideal for PRs.

With your branch at 08279e1db7e435df2bbc0dd05f83736c5e6a5b1e, you should be able to do
```
git rebase --onto main 7ee9643e7ff1633c709f4ffd683050eb3c726ec9 fix-transition-order
```

Where 7ee9643e7ff1633c709f4ffd683050eb3c726ec9 is the nearest common ancestor of your branch and https://github.com/servo/stylo/commits/2025-03-11/

yezhizhen:
Somehow I cannot compile anymore..

I am on the latest Servo main branch. 
Latest Stylo branch, with my five commits.
But I get a lot of 

```
error[E0053]: method `report_error` has an incompatible type for trait
   --> components\script\dom\fontface.rs:180:20
    |
180 |         _location: cssparser::SourceLocation,
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^ expected `style::values::SourceLocation`, found `cssparser::SourceLocation`
```

nicoburns:
> Somehow I cannot compile anymore.. I just fetched the latest Servo main branch.
> I am on the latest Servo main branch. Latest Stylo branch, with my five commits.

@yezhizhen You upgraded to the latest Stylo branch a little too quickly! CI is still running on the corresponding Servo PR (https://github.com/servo/servo/pull/35990). Your code should compile again once that merges. Until then you can run against the `2025-03-12` Stylo branch.

yezhizhen:
Haha thanks Nico! I will get back tomorrow then :)

yezhizhen:
I still cannot compile locally, with latest code
I think there is something wrong with current `Cargo.toml` for using local stylo.

I made some adjustments. It starts to compile, but has following errors now:

```
error[E0277]: the trait bound `ElementState: MallocSizeOf` is not satisfied
   --> D:\Rust\Cargo\git\checkouts\stylo-482338307e42a9ea\6aa5733\style\servo\selector_parser.rs:778:26
    |
778 | #[derive(Debug, Default, MallocSizeOf)]
    |                          ^^^^^^^^^^^^ the trait `MallocSizeOf` is not implemented for `ElementState`
```

yezhizhen:
https://github.com/yezhizhen/servo/actions/runs/13914368346/job/38936024866

Failing WPT test after optimization "Revamp".. Fixing it..

yezhizhen:
https://github.com/yezhizhen/servo/actions/runs/13918080266/job/38947114999

I reverted the "optimization commits", and ran same tests as yesterday again. But somehow, it is also failing. Previously, it always pass.

Is it because something changed, after recent Stylo rebase?

Because I tested those two failing tests locally, which always has expected behaviour.

yezhizhen:
Passing locally already. 
```
./mach test-wpt _mozilla/css/transitionend_event.html _mozilla/css/transition_calc_implicit.html

Running WPT tests with D:\servo\target\debug\servo.exe
Running 2 tests in web-platform-tests

Ran 2 tests finished in 5.8 seconds.
  • 2 ran as expected.
```

Running Git actions now. 

https://github.com/yezhizhen/servo/actions/runs/13919914249

style/animation.rs
```diff
@@ -1012,101 +987,140 @@ impl ElementAnimationSet {
...
 
-        if let Some(old_transition) = self
-            .transitions
-            .iter_mut()
-            .filter(|transition| transition.state == AnimationState::Running)
-            .find(|transition| {
-                transition.property_animation.property_id() == *property_declaration_id
-            })
-        {
             // We always cancel any running transitions for the same property.
```
stevennovaryo, line 1115: I suppose this is the effect of condition 4. Could we mention it here?

style/animation.rs
```diff
@@ -1012,101 +987,140 @@ impl ElementAnimationSet {
...
+                } 
+            } else{
+                start_new_transition = false;
+            }
+            running_transition = Some(old_transition);
         }
 
+        // Step 1 + 4.3 + 4.4 in https://drafts.csswg.org/css-transitions/#starting
         // We are going to start a new transition, but we might have to update
         // it if we are replacing a reversed transition.
```
stevennovaryo, line 1095:
```suggestion
        // We are going to start a new transition, but we might have to update
        // it if we are replacing a reversed transition for step 4.3.
```

style/animation.rs
```diff
@@ -1012,101 +987,140 @@ impl ElementAnimationSet {
...
+            .iter_mut()
             .filter(|transition| transition.state != AnimationState::Canceled)
-            .any(|transition| transition.property_animation.to == property_animation.to)
+            .find(|transition| {
+                transition.property_animation.property_id() == property_declaration
+            })
         {
-            return;
+            // and the end value of the running transition is not equal to the value of the property in the after-change style, then:"
+            if to != old_transition.property_animation.to {
```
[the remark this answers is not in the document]
yezhizhen: It is already checked immediately after this step. 

We are just checking if it is possible to start new transition here.
Later, we are searching `running/pending` transition to cancel them.

In the context of Servo, `pending` also means the word "running" in the spec.

style/animation.rs
```diff
@@ -1012,101 +987,140 @@ impl ElementAnimationSet {
...
+                                            (!allow_discrete && !current_val.interpolable_with(&to));
+
+                // Step 4.1
+                //"If the current value of the property in the running transition is equal 
+                // to the value of the property in the after-change style, 
+                // or if these two values are not transitionable, then implementations must cancel the running transition."
+                if current_val == to || not_transitionable {
+                    old_transition.state = AnimationState::Canceled;
+                    self.dirty = true;
+                    start_new_transition = false;
```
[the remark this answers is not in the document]
yezhizhen: No. It is still possible to cancel more tansitions in the next if branch.

style/animation.rs
```diff
@@ -1012,101 +987,140 @@ impl ElementAnimationSet {
...
+                // "Otherwise, if the combined duration is less than or equal to 0s, or if the current value of the property in the 
+                // running transition is not transitionable with the value of the property in the after-change style,
+                // then implementations must cancel the running transition."
+                else if duration + delay <= 0.0 {
+                    old_transition.state = AnimationState::Canceled;
+                    self.dirty = true;
+                    start_new_transition = false;
+                } 
+            } else{
+                start_new_transition = false;
```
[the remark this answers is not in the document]
yezhizhen: Because here the `else` means the running transition already have same `to` as `after_change_to`. So we don't start.

style/animation.rs
```diff
@@ -1012,101 +987,140 @@ impl ElementAnimationSet {
...
+            }
+            running_transition = Some(old_transition);
```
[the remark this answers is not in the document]
yezhizhen: It is already one of `Pending`, `Paused`, `Running`, which means "running transition".

yezhizhen:
I just added a new wpt-test in https://github.com/servo/servo/pull/35978, which Servo was failing before this PR.

But the Servo PR has no reviewers yet.

style/animation.rs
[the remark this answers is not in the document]
yezhizhen: Thanks!

style/animation.rs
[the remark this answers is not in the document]
yezhizhen: Thanks!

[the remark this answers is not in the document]
yezhizhen: Thanks! I've adjusted accordingly in the latest commit.

mrobinson (changes requested):
This change greatly modifies the structure of the code. The previous code was a lot easier to reason about, so it is difficult to review this version. Can you please try to preserve the original structure as much as possible and simply modify the bits that need to be changed? Thanks.

yezhizhen:
Yes, but that would just last three commits involving optimizations, and computationally more heavy to find the animations to cancel.

Should I just remove these optimizations? @mrobinson

yezhizhen:
Yes, I can just last three commits involving optimizations, and original structure won't be affected much.
Should I just remove these optimizations? @mrobinson 

Or, should I just apply your requested changes?
Thanks for the review!

mrobinson:
> Or, should I just apply your requested changes, while keeping rest as close to original structure as possible? (I will work towards **this direction** for now.)

This sounds good. Thanks for being adaptable! Yeah, the main issue is that it is hard to pinpoint exactly what was the issue with the original code because the structure has changed a bunch. I think by preserving some of the original flow, but still fixing the issues (as much as possible), it will be easier to see what you have fixed. Thank you!

style/animation.rs
```diff
@@ -1012,101 +987,147 @@ impl ElementAnimationSet {
...
-        }
+        let not_transitionable = !property_declaration.is_animatable() || 
+                                        (!allow_discrete && property_declaration.is_discrete_animatable());
+        
+        let mut start_new_transition = !not_transitionable;
 
         let timing_function = style.transition_timing_function_mod(index);
-        let duration = style.transition_duration_mod(index);
+        let duration = style.transition_duration_mod(index).seconds() as f64;
         let delay = style.transition_delay_mod(index).seconds() as f64;
```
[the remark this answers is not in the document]
yezhizhen: But all we need is its seconds value tho; and it is keep reused in later context.

style/animation.rs
```diff
@@ -1012,101 +987,147 @@ impl ElementAnimationSet {
...
+        }
+
+        // FIXME(emilio): Handle the case where old_style and new_style's writing mode differ.
+        let Some(from) = AnimationValue::from_computed_values(property_declaration, old_style) else {
+            return;
+        };
+        let Some(to) = AnimationValue::from_computed_values(property_declaration, new_style) else {
+            return;
+        };
+
```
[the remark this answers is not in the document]
yezhizhen: The old function is `from_property_declaration` was only used once. I find it much more straightforward, inlining it here..

style/animation.rs
```diff
@@ -1012,101 +987,147 @@ impl ElementAnimationSet {
...
             .find(|transition| {
-                transition.property_animation.property_id() == *property_declaration_id
+                transition.property_animation.property_id() == property_declaration
             })
         {
-            // We always cancel any running transitions for the same property.
-            old_transition.state = AnimationState::Canceled;
-            new_transition.update_for_possibly_reversed_transition(old_transition, delay, now);
+            // and the end value of the running transition is not equal to the value of the property in the after-change style, then:"
+            if to != old_transition.property_animation.to {
```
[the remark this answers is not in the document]
yezhizhen: We can't early return; we will return after we cancel the `old_transition`!
Which is why previous code fails to cancel animations. 
I have written a new wpt-test case for this.

https://github.com/servo/servo/pull/35978/files#diff-382b686e6201a7b1eb53044460d46d845f01f7d515326b3028564e1ac1f0addc

style/animation.rs
```diff
@@ -1012,101 +987,147 @@ impl ElementAnimationSet {
...
+                    }
+                    running_transition = Some(old_transition);
+                } 
+            } 
+            // Per [1], don't trigger a new transition if the end state for that
+            // transition is the same as that of a transition that's running or
+            // completed. We don't take into account any canceled animations.
+            // [1]: https://drafts.csswg.org/css-transitions/#starting
+            else {
+                start_new_transition = false;
```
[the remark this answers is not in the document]
yezhizhen: Changing now. I was trying to keep a more uniform state tracker.

style/animation.rs
```diff
@@ -1129,6 +1150,26 @@ impl ElementAnimationSet {
...
+        for transition in &self.transitions {
+            if transition.state == AnimationState::Canceled || transition.state == AnimationState::Finished {
+                continue;
+            }
+            let value = transition.calculate_value(now);
+            map.insert(value.id().to_owned(), value);
+        }
+
+        Some(map)
+    }
```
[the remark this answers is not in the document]
yezhizhen: I was thinking about same; but felt a bit awkward to create a enum just for this so I didn't do this before :)

yezhizhen:
> We can't early return; we will return after we cancel the `old_transition` if appropriate! Which is **_why previous code fails to cancel many animations_**. Also why we need to restructure the code a bit.
> 
> I have written a new wpt-test case for this, which the previous code would [FAIL].
> 
> https://github.com/servo/servo/pull/35978/files#diff-382b686e6201a7b1eb53044460d46d845f01f7d515326b3028564e1ac1f0addc

In case other reviewers are wondering, this is the key idea about cancellation in this PR; also why the structure of the code needs to be changed a bit.

style/animation.rs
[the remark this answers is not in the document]
yezhizhen: Also, by using `seconds` directly, we can remove the import of `Time`.

yezhizhen:
For example, previous code would simply return early if `duration < 0`. But even if we won't start new transitions, we still need to cancel running transitions.

yezhizhen:
> it will be easier to see what you have fixed. Thank you!

https://github.com/user-attachments/assets/c571d623-cdf3-4870-8e15-9e8da1854d04


https://github.com/user-attachments/assets/4f6407e6-3837-4295-b393-52ba7fe04187

style/animation.rs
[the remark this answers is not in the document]
yezhizhen: I am willing to explain in video call too.

mrobinson:
Okay. I've spent some time looking more closely at this change and I think it's a good start! Apologies for the current mess, but I think this is a good opportunity to actually implement the specification in the order it is written.

1. Add a `start_new_transition` method that can be used for early returns.
2. Restructure `start_transition_if_applicable` to match https://web.archive.org/web/20250322011556/https://drafts.csswg.org/css-transitions/#starting.
   a. First check all the conditions of step 1, caching any values that are used twice, such as:
       ```rust
           let combined_duration_is_greater_than_zero = ...
       ```
   b. If all the conditions for 1 are met, you can do an early return:
       ```rust
           self.start_new_transition(...);
           return;
       ```
   c. After step 1 you can do an early return if there is no pre-existing transition as all the subsequent steps rely on there being one:
       ```rust
            if let Some(existing_transition) = existing_transition else {
                return;
            }
       ```
3. The final part of `start_transition_if_applicable` would be checking the various specification conditions related to `existing_transition` in the order specified, finally canceling `existing_transition` and starting the new transition in step 4 if necessary.

yezhizhen:
> Okay. I've spent some time looking more closely at this change and I think it's a good start! Apologies for the current mess, but I think this is a good opportunity to actually implement the specification in the order it is written.

Thanks for the suggestions! When I opened this PR I try to work around original code without altering it too much. It is nice we can do this now :)

yezhizhen:
@mrobinson I just restructured the code to follow exact same order as spec. And my newly added wpt-tests and other manual tests are still passing.

mrobinson:
@yezhizhen I've pushed a couple commits onto your branch with some minor changes. Please take a look. Do these seeem okay to you?

style/animation.rs
```diff
@@ -990,187 +1015,232 @@ impl ElementAnimationSet {
...
+        // > If all of the following are true:
+        // > - the element does not have a running transition for the property,
+        // > - the before-change style is different from the after-change style for that property,
+        // >   and the values for the property are transitionable,
+        // > - there is a matching transition-property value, and
+        // > - the combined duration is greater than 0s,
+        //
+        // This function is only run if ther eis a matching transition-property
+        // value, so that check is skipped here.
+        let has_running_transition = existing_transition.as_ref().is_some_and(|transition| {
```
yezhizhen, line 1088: Typo: there is
Also, the check is already done and matched; otherwise we won't enter `fn start_transition_if_applicable`

style/animation.rs
```diff
@@ -990,187 +1015,232 @@ impl ElementAnimationSet {
...
-        // > 2. the before-change style is different from the after-change style for that property, 
-        // >    and the values for the property are transitionable,
-        // > 4. there is a matching transition-property value, and
-        // > 5. the combined duration is greater than 0s,
-        if !has_running_transition && (transitionable && from != to) && (duration + delay > 0.0) {
-            // > 3. the element does not have a completed transition for the property or the end value 
-            // >    of the completed transition is different from the after-change style for the property
-            let mut start_new_transition = !has_completed_transition;
-            if has_completed_transition {
-                let completed_transition = old_transition.as_ref().unwrap();
```
yezhizhen: This condition 3 also needs to be checked in step 1.
Right now it is removed in your latest commit.

style/animation.rs
```diff
@@ -990,187 +1015,232 @@ impl ElementAnimationSet {
...
-                    reversing_shortening_factor: 1.0,
-                };
-                self.transitions.push(new_transition);
+
+            // > Step 4.2: Otherwise, if the combined duration is less than or
+            // > equal to 0s, or if the current value of the property in the
+            // > running transition is not transitionable with the value of the
+            // > property in the after-change style, then implementations must
+            // > cancel the running transition.
+            if duration + delay < 0.0 {
```
yezhizhen, line 1164: Should be `duration + delay <= 0.0`

style/animation.rs
```diff
@@ -990,187 +1015,232 @@ impl ElementAnimationSet {
...
             return None;
         }
 
         let mut map =
             AnimationValueMap::with_capacity_and_hasher(self.transitions.len(), Default::default());
         for transition in &self.transitions {
-            match flag {
-                IncludedTransitions::Active => {
+            match ignore_transitions {
+                IgnoreTransitions::Canceled => {
```
yezhizhen, line 1237: I feel a bit clumsy, as this check is done in each iteration despite the static execution path. And we need to introduce a extra enum.

Does compiler optimize it away? Would it still be better, if we just create two functions?

style/matching.rs
```diff
@@ -605,7 +605,7 @@ trait PrivateMatchMethods: TElement {
     fn process_animations_for_pseudo(
         &self,
         context: &mut StyleContext<Self>,
-        old_styles: &ElementStyles,
+        old_styles: &mut ElementStyles,
```
yezhizhen: It is not mutated tho. Would it be better to just remove `mut`?

style/matching.rs
```diff
@@ -687,8 +687,8 @@ trait PrivateMatchMethods: TElement {
     fn process_animations_for_style(
         &self,
         context: &mut StyleContext<Self>,
-        old_values: &Option<Arc<ComputedValues>>,
-        new_values: &Arc<ComputedValues>,
+        old_values: &mut Option<Arc<ComputedValues>>,
+        new_values: &mut Arc<ComputedValues>,
```
yezhizhen: Ditto.

yezhizhen (commented):
@mrobinson  LGTM mostly, but Condition 3 of step 1 is not checked in your latest commit.
I also have some other questions below :)

style/animation.rs
yezhizhen: This condition 3 also needs to be checked in step 1.
Right now it is removed in your latest commit.

style/animation.rs
```diff
@@ -990,187 +1015,232 @@ impl ElementAnimationSet {
...
-        // >    and the values for the property are transitionable,
-        // > 4. there is a matching transition-property value, and
-        // > 5. the combined duration is greater than 0s,
-        if !has_running_transition && (transitionable && from != to) && (duration + delay > 0.0) {
-            // > 3. the element does not have a completed transition for the property or the end value 
-            // >    of the completed transition is different from the after-change style for the property
-            let mut start_new_transition = !has_completed_transition;
-            if has_completed_transition {
-                let completed_transition = old_transition.as_ref().unwrap();
-                start_new_transition = completed_transition.property_animation.to != to;
```
yezhizhen: This condition 3 also needs to be checked in step 1.
Right now it is removed in your latest commit.

yezhizhen:
https://github.com/yezhizhen/servo/actions/runs/14217536133

@mrobinson The latest try run in Servo PR for this Stylo PR is successful.
Also, what do you think about above unresolved conversation?

mrobinson:
@yezhizhen Sorry, for some reason my comment on the unresolved conversation was left in a draft state. You can find it above now.

mrobinson:
If this looks okay to you, I will squash the changes and land the PR.

yezhizhen:
@mrobinson LGTM. Please land :)

yezhizhen:
@mrobinson LGTM. Please land :)

For the Servo PR, I believe that I need to rebase and revert `Cargo.toml` and `Cargo.lock`?

mrobinson (approved):
Apologies for the long process to review this!

mrobinson:
I'm going to land this one event though the Servo PR isn't approved yet as it shouldn't cause any issues in Servo except another test will start passing.

yezhizhen:
> I'm going to land this one event though the Servo PR isn't approved yet as it shouldn't cause any issues in Servo except another test will start passing.

Should be fine as `Cargo.lock` in Servo main branch hasn't updated yet.

PR #149, opened 2025-03-17
Make servo feature default

nicoburns:
We probably won't be able to upstream this. But IMO it would be reasonable for us to keep this downstream. And it's a nice QoL improvement for users of the crate (and indeed for us when we want to build Stylo standalone).


Loirooriol (approved):
Yeah having to type `--features servo` is annoying, thanks

PR #152, opened 2025-03-17
Revert renaming of generic parameter

nicoburns:
This is aimed at reducing the diff. We could upstream this, but as this is a non-functional change (and I believe most of the rest of the generic values use this style of long generic names), I propose we just adjust this back to match upstream.

FWIW, the long generic names confused me for a long time: why is it called `LengthPercentage` when it could really be anything? What I realised is that the "generic" values are designed to be generic specifically over the "specified" and "computed" values. So the intent of this code is the `LengthPercentage` parameter is either `values::specified::LengthPercentage` or `values::computed::LengthPercentage`. With that in mind, the `LengthPercentage` name makes more sense I think.

### DIFF
--- a/style/values/generics/length.rs
+++ b/style/values/generics/length.rs
@@ -75,13 +75,13 @@ impl<LengthPercentage> LengthPercentageOrAuto<LengthPercentage> {
     }
 }
 
-impl<T> LengthPercentageOrAuto<T>
+impl<LengthPercentage> LengthPercentageOrAuto<LengthPercentage>
 where
-    T: Clone,
+    LengthPercentage: Clone,
 {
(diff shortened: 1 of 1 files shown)


Loirooriol (approved):
My recollection was that this was needed to allow mapping between two different types `LengthPercentageOrAuto<T>` and `LengthPercentageOrAuto<U>` where one of T or U would be `Au` and the other an actual `<length-percentage>`.

But upstream already uses `pub fn map<T>(&self, f: impl FnOnce(LengthPercentage) -> T) -> LengthPercentageOrAuto<T> {`, with `T` being possibly different than `LengthPercentage`.

So yeah, it seems this change was unnecessary.

PR #156, opened 2025-03-23
Replace fxhash with rustc-hash

paolobarbolini:
The `fxhash` crate hasn't been maintained in the last 6 years. This migrates to `rustc-hash`, which is much more popular crate developed by the Rust project and used by the Rust compiler.


mrobinson:
@paolobarbolini Thanks for the contribution.

This crate is a downstream version of Gecko's style. We ask that non-Servo-specific change be submitted upstream: https://searchfox.org/mozilla-central/source/servo/components.

nicoburns:
There's a guide to submitting patches to Gecko here: https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html

However, I would note that "fxhash" stands for Firefox Hash so I'm not sure how open they would be to this change.

Loirooriol:

Though as mentioned above, this should probably be sent upstream instead.

PR #163, opened 2025-03-31
Update names of lib targets

nicoburns:
While the *package* names of the crates have all been updated to stylo*, it seems like that name of the *crate within the package* is still style* in some cases. This causes the docs.rs url to be https://docs.rs/stylo/latest/style rather than https://docs.rs/stylo/latest/stylo.

This PR changes this to be consistent by simply deleting the `name = "..."` lines which causes the crate name to default to the package name.

It also removes the `crate_name` attribute from the `stylo_traits` crate. These need to match the crate name in `Cargo.toml`. And as upstream have not yet renamed this crate keeping this line would mean extra diff. This line comes from the very first version of `style_traits` in 2015 (https://github.com/servo/stylo/commit/d7ba5ee71f669412697bb0caaeea11182b216b80) and none of the other crates have this attribute so I strongly suspect it is not necessary.


style_traits/lib.rs
```diff
@@ -5,8 +5,6 @@
 //! This module contains shared types and messages for use by devtools/script.
 //! The traits are here instead of in script so that the devtools crate can be
 //! modified independently of the rest of Servo.
-
-#![crate_name = "style_traits"]
```
Loirooriol, line 9: https://doc.rust-lang.org/rust-by-example/attribute/crate.html

> it is important to note that both the crate_type and crate_name attributes have no effect whatsoever when using Cargo

So was this added in case that cargo is not used??

I'm a bit worried about accidentally breaking something that I'm missing.

nicoburns: > So was this added in case that cargo is not used??

I think so. See: https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/rust-by-example/attribute/crate.html Given the timing of this code (when Cargo was still only a few months old), I suspect it may have been written with a pre-cargo world in mind.

nicoburns: This passes Mozilla CI (https://treeherder.mozilla.org/jobs?repo=try&revision=a0ba50e438a0bdedc8e36fa026fe9550d6e3dd71). Perhaps I should just submit this upstream?

Loirooriol: Yes submitting it upstream sounds better

Loirooriol:

PR #165, opened 2025-04-01
Only set PYTHONDONTWRITEBYTECODE env variable for subprocess

nicoburns:
`env::set_var` has become `unsafe` in the 2024 edition (https://github.com/rust-lang/rust/issues/124866), so best that we avoid it.

### DIFF
--- a/style/build.rs
+++ b/style/build.rs
@@ -53,18 +53,17 @@ fn generate_properties(engine: &str) {
         }
     }
 
-    // `cargo publish` isn't happy with the `__pycache__` files that are created
-    // when we run the property generator.
-    //
-    // TODO(mrobinson): Is this happening because of how we run this script? It
-    // would be better to ensure are just placed in the output directory.
-    env::set_var("PYTHONDONTWRITEBYTECODE", "1");
-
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

style/build_gecko.rs
```diff
@@ -369,6 +369,7 @@ fn generate_atoms() {
         .join("regen_atoms.py");
     println!("cargo:rerun-if-changed={}", script.display());
     let status = Command::new(&*PYTHON)
+        .env("PYTHONDONTWRITEBYTECODE", "1")
```
mrobinson, line 372: I don't think this one is necessary because `build_gecko::generate()` isn't run when we can `cargo publish`.

mrobinson (changes requested):
Just a small change request:

mrobinson (approved):
Sorry, but with that change this is good. Thanks!

style/build_gecko.rs
nicoburns: Popped that in there for equivalence with the previous code. But yeah, it can probably go.

PR #168, opened 2025-04-06
Box `PaintWorklet` to reduce `GenericImage` size

### LINKED ISSUE #146: Reduce size of `GenericImage`
Gecko's version of `GenericImage` is 16 bytes. Servo's is 40 bytes.

As images are used in quite a lot of places, keeping the size small is important for memory usage.

We should reduce the size of our version of `GenericImage` by boxing the `PaintWorklet` variant.

sidntrivedi012:
Closes #146. Reduces the size of `GenericImage` from 40 to 24 bytes by boxing `PaintWorklet` variant.

### DIFF
--- a/style/properties/longhands/border.mako.rs
+++ b/style/properties/longhands/border.mako.rs
@@ -112,7 +112,6 @@ ${helpers.predefined_type(
     spec="https://drafts.csswg.org/css-backgrounds/#the-background-image",
     vector=False,
     animation_type="discrete",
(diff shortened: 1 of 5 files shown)


nicoburns (approved):

nicoburns:
Looks good. I see we're still not quite as small as Gecko, and I've opened https://github.com/servo/stylo/issues/169 to track taht.

PR #170, opened 2025-04-06
Reduce size of `CssUrl`

### LINKED ISSUE #169: Reduce size of `CssUrl`
As a followup to #146, we should action the following TODO:

```rust
/// TODO(emilio): This should be shrunk by making CssUrl a wrapper type of an
/// arc, and keep the serialization in that Arc. See gecko/url.rs for example.
```

by updating the Servo version of `CssUrl` (https://github.com/servo/stylo/blob/main/style/servo/url.rs#L26) to be a newtype of a single Arc<T> (where T is a new `CssUrlData` struct) like the Gecko version of `CssUrl` (https://github.com/servo/stylo/blob/main/style/gecko/url.rs#L27).

This will bring `CssUrl` and thus `Image` down to 16 bytes from 24.

sidntrivedi012:
Closes #169. Reduces the size of `CssUrl` and hence, the size of `Image` for servo from 24 to 16 bytes.

### DIFF
--- a/style/servo/url.rs
+++ b/style/servo/url.rs
@@ -12,6 +12,7 @@ use servo_arc::Arc;
 use std::fmt::{self, Write};
 use style_traits::{CssWriter, ParseError, ToCss};
 use to_shmem::{SharedMemoryBuilder, ToShmem};
+use std::ops::Deref;
 use url::Url;
 
 /// A CSS url() value for servo.
@@ -22,11 +23,16 @@ use url::Url;
 ///
 /// However, this approach is still not necessarily optimal: See
 /// <https://bugzilla.mozilla.org/show_bug.cgi?id=1347435#c6>
-///
-/// TODO(emilio): This should be shrunk by making CssUrl a wrapper type of an
-/// arc, and keep the serialization in that Arc. See gecko/url.rs for example.
 #[derive(Clone, Debug, Deserialize, MallocSizeOf, Serialize, SpecifiedValueInfo)]
-pub struct CssUrl {
+#[css(function = "url")]
+#[repr(C)]
+pub struct CssUrl(#[ignore_malloc_size_of = "Arc"]pub Arc<CssUrlData>);
+
+/// Data shared between CssUrls.
+///
+#[derive(Debug, Deserialize, MallocSizeOf, Serialize, SpecifiedValueInfo)]
+#[repr(C)]
+pub struct CssUrlData {
     /// The original URI. This might be optional since we may insert computed
     /// values of images into the cascade directly, and we don't bother to
     /// convert their serialization.
@@ -47,6 +53,13 @@ impl ToShmem for CssUrl {
     }
 }
 
+impl Deref for CssUrl {
+    type Target = CssUrlData;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 impl CssUrl {
     /// Try to parse a URL from a string value that is a valid CSS token for a
     /// URL.
@@ -55,10 +68,10 @@ impl CssUrl {
     pub fn parse_from_string(url: String, context: &ParserContext, _: CorsMode) -> Self {
         let serialization = Arc::new(url);
         let resolved = context.url_data.0.join(&serialization).ok().map(Arc::new);
-        CssUrl {
-            original: Some(serialization),
-            resolved: resolved,
-        }
+        CssUrl(Arc::new(CssUrlData {
+             original: Some(serialization),
+             resolved: resolved 
+            }))
     }
 
     /// Returns true if the URL is definitely invalid. For Servo URLs, we can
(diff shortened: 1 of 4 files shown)


style/values/generics/image.rs
```diff
@@ -28,7 +28,7 @@ pub enum GenericImage<G, ImageUrl, Color, Percentage, Resolution> {
     /// `none` variant.
     None,
     /// A `<url()>` image.
-    Url(ImageUrl),
+    Url(Box<ImageUrl>),
```
nicoburns, line 31: I don't think this should be boxed. `ImageUrl` (which is a type alias for `CssUrl`) already contains an `Arc`, so boxing here is double boxing.

nicoburns (changes requested):

style/values/generics/image.rs
sidntrivedi012: @nicoburns have removed boxing from here but the computed image size still remains at 24 bytes. The specified image size got reduced to 16 though. Any pointers as to why that would be happening?

style/values/specified/image.rs
```diff
@@ -46,7 +46,7 @@ pub type Image = generic::Image<Gradient, SpecifiedUrl, Color, Percentage, Resol
 #[cfg(feature = "gecko")]
 size_of_test!(Image, 16);
 #[cfg(feature = "servo")]
-size_of_test!(Image, 24);
+size_of_test!(Image, 16);
```
Loirooriol, line 49: You can unify these since both Gecko and Servo are 16 now.

style/servo/url.rs
```diff
@@ -55,10 +68,10 @@ impl CssUrl {
...
         let serialization = Arc::new(url);
         let resolved = context.url_data.0.join(&serialization).ok().map(Arc::new);
-        CssUrl {
-            original: Some(serialization),
-            resolved: resolved,
-        }
+        CssUrl(Arc::new(CssUrlData {
+             original: Some(serialization),
+             resolved: resolved 
+            }))
```
Loirooriol, line 74: Don't mess indentation, and don't replace the comma with a trailing space.

Loirooriol (changes requested):

style/values/generics/image.rs
nicoburns: This seems to be because `ComputedUrl` (https://github.com/servo/stylo/blob/main/style/servo/url.rs#L209) is still 16 bytes (8 + enum tag) which becomes 24 when combined with the tag for `Image`.

style/values/specified/image.rs
sidntrivedi012: Done.

style/servo/url.rs
sidntrivedi012: Made the fix. Can you share what formatter should I be using in my dev environment? As of now, I am using default rust-analyzer.

style/values/generics/image.rs
sidntrivedi012: Oh, got it. So, should we aim to get it to 16 in this PR or should we take it up in another one?

sidntrivedi012:
@Loirooriol @nicoburns can you please take another look at the PR? Have addressed the review comments.

style/values/specified/image.rs
```diff
@@ -43,10 +43,9 @@ fn gradient_color_interpolation_method_enabled() -> bool {
 pub type Image = generic::Image<Gradient, SpecifiedUrl, Color, Percentage, Resolution>;
 
 // Images should remain small, see https://github.com/servo/servo/pull/18430
-#[cfg(feature = "gecko")]
+#[cfg(any(feature = "gecko", feature = "servo"))]
```
Loirooriol, line 46: You can just remove the cfg since there is no other possibility than servo and gecko.

style/servo/url.rs
Loirooriol: Use `rustfmt style/servo/url.rs`

style/values/specified/image.rs
```diff
@@ -43,10 +43,9 @@ fn gradient_color_interpolation_method_enabled() -> bool {
...
 size_of_test!(Image, 16);
-#[cfg(feature = "servo")]
-size_of_test!(Image, 24);
+
```
Loirooriol, line 48: Remove extra newline

style/servo/url.rs
Loirooriol, line 1: Fix formatting with `rustfmt style/servo/url.rs`

nicoburns: @Loirooriol I thought we didn't auto-format Stylo? Is there a way to automatically format the whole repo to match the current formatting? If I run `cargo +nightly fmt` there are 94 files changed...

Loirooriol: No, it's not formatted in general, but that's not a reason to break the formatting here. And `style/servo/url.rs` is properly formatted anyways.

If you want to format the entire thing you can do it upstream. Mozilla does it from time to time, e.g. https://phabricator.services.mozilla.com/D158234

style/values/specified/image.rs
sidntrivedi012: Removed.

style/values/specified/image.rs
sidntrivedi012: Done.

style/servo/url.rs
sidntrivedi012: Ran `rustfmt` on `style/servo/url.rs`.

nicoburns (approved):
I'll let Oriol review this too as he previously requested changes, but LGTM. I think we can defer bringing the ComputedUrl down to 16 bytes to a later PR as I'm not 100% sure what the best approach for that will be whereas this seem uncontroversial.

sidntrivedi012:
@nicoburns can you share any other issue as well that I can take up?

Loirooriol (approved):
Looks good, but before landing please open a PR in Servo that overrides the Stylo crates to this PR, to confirm that everything is fine.

sidntrivedi012:
@nicoburns @Loirooriol the following check in servo has passed: https://github.com/servo/servo/actions/runs/14474436889/job/40597077575?pr=36545.

nicoburns:
Servo checks are passing

PR #175, opened 2025-04-29
Enable `font-feature-settings` longhand property

nicoburns:
This property allows font features such as ligatures and smallcaps to be enabled.


Loirooriol:

PR #179, opened 2025-05-05
Add release information to README and make sure it's used for crates

mrobinson:


nicoburns (approved):
Arguably might be better in `SYNCING.MD` as it's really maintainer-oriented rather than user-oriented information (although release frequency could be useful info for users). But I think it's also fine here.

mrobinson:
I plan to make some changes here, but right now this should reflect what was published for `0.3.0`.

PR #180, opened 2025-05-12
Enable moz-block-height

stevennovaryo:
Enable -moz-block-height, internal value for CSS line-height property that is only usable in UA stylesheet.

Specifically, it will resolve the line height calculation to the containing block height, and in the calculation of the block height contribution, it will resolve like line-height: normal.

This property would then be able to be used to correct <input> element widget as specified by https://html.spec.whatwg.org/multipage/rendering.html#the-input-element-as-a-text-entry-widget, where the used value for it's line height should be greater or equal to normal. And, https://drafts.csswg.org/css-ui/#input-rules where by default, one line text control should have it's content be aligned center vertically. Though further encapsulation with UA shadow dom is still necessary.

Stylo side of: [36930](https://github.com/servo/servo/pull/36930)


style/servo/media_queries.rs
```diff
@@ -180,6 +180,8 @@ impl Device {
             LineHeight::Normal => CSSPixelLength::new(0.),
             LineHeight::Number(number) => font.font_size.computed_size() * number.0,
             LineHeight::Length(length) => length.0,
+            // TODO: this should resolve like `normal` too.
+            LineHeight::MozBlockHeight => CSSPixelLength::new(0.),
```
Loirooriol, line 184:
```suggestion
            // TODO: compute `normal` and `-moz-block-height` from the font metrics.
            LineHeight::Normal | LineHeight::MozBlockHeight => CSSPixelLength::new(0.),
            LineHeight::Number(number) => font.font_size.computed_size() * number.0,
            LineHeight::Length(length) => length.0,
```

stevennovaryo: Done, thanks!

Loirooriol:

PR #183, opened 2025-05-15
Make more CSS properties only require a repaint for Servo

mrobinson:


style/properties/longhands/box.mako.rs
```diff
@@ -435,6 +435,7 @@ ${helpers.single_keyword(
     spec="https://drafts.csswg.org/css-transforms/#backface-visibility-property",
     extra_prefixes=transform_extra_prefixes,
     animation_type="discrete",
+    servo_restyle_damage="repaint",
```
Loirooriol, line 438: `backface-visibility` can establish a containing block for all descendants: https://drafts.csswg.org/css-transforms-2/#backface-visibility-property

style/properties/longhands/inherited_box.mako.rs
```diff
@@ -12,6 +12,7 @@ ${helpers.single_keyword(
     gecko_ffi_name="mVisible",
     spec="https://drafts.csswg.org/css-box/#propdef-visibility",
     gecko_enum_prefix="StyleVisibility",
+    servo_restyle_damage="repaint",
```
Loirooriol, line 15: `visibility` can establish an independent formatting context

nicoburns:
It seems like this is treating each property individually and setting damage if the property *may* change whether the node establishes a containing block, positioning context, independent formatting context, etc.

It would be nice if for each of "containing block", "positioning context", "independent formatting context" we:
- Check all properties which can effect this so that we fully determine whether it establishes a new context or not
- Store the result of this such that we can compare with the result from the last style run
- Only mark damage if it has actually changed.

Perhaps as an intermediary step these could at least become separate categories of damage?

style/properties/longhands/inherited_box.mako.rs
Loirooriol: Actually, no, it can't. I think I saw the `visible` and `hidden` values and I thought about `overflow`.

But anyways, `visibility: collapse` can affect layout.

Loirooriol (approved):

PR #186, opened 2025-05-20
Use firefox repo as upstream

### LINKED ISSUE #184: Switch upstream to https://github.com/mozilla-firefox/firefox
Mozilla has a new official git repository for Firefox (/Gecko) (https://github.com/mozilla-firefox/firefox) which replaces the mercurial repositories as the canonical source of truth for Mozilla code. We should switch our upstream to point to this.

nicoburns:
- Fixes #184

Mozilla have deprecated https://github.com/mozilla/gecko-dev in favour of https://github.com/mozilla-firefox/firefox. This PR switches our syncing code over to the new repo.

A one-off manual deletion of local `_cache` and `_filtered` directories will be required to use this PR.

I have run the sync script in this PR and pushed the result to the [ffupstream](https://github.com/servo/stylo/tree/ffupstream) branch for reference.

### DIFF
--- a/SYNCING.md
+++ b/SYNCING.md
@@ -19,7 +19,7 @@ $ rm -Rf _filtered
 Now overwrite our `upstream` with those commits and push:
 
 ```sh
-$ git fetch -f --progress ./_filtered master:upstream
+$ git fetch -f --progress ./_filtered main:upstream
 $ git push -fu --progress origin upstream
 ```
 
--- a/sync.sh
+++ b/sync.sh
@@ -30,7 +30,7 @@ fi
 
 step Cloning upstream if needed
 if ! [ -e upstream ]; then
-    git clone --bare --single-branch --progress https://github.com/mozilla/gecko-dev.git upstream
+    git clone --bare --single-branch --progress https://github.com/mozilla-firefox/firefox.git upstream
 fi
 
 step Updating upstream


mrobinson:
@nicoburns It looks like hashes are not the same between upstreams, so I suppose we should decide what we want to do here. I think there are a few options:

1. Duplicate the entire repository contents and history.
2. Rewrite history so that existing branches share hashes with the new upstream.
3. Keep investigating to see if there is a way to ensure that the hashes are the same.

jschwe:
We could rename relevant existing branches (e.g. to `old_xxx`), so that the history and existing commit hashes stay preserved.

nicoburns:
Using [DirEqual](https://naarakstudio.com/direqual/) I have been able to confirm that the synced versions of the old and the new repos have identical contents. Methodology:
- Checkout https://github.com/servo/stylo/commit/1107f7a2cac4c404bc10884b49de1daed0433ddb
- Checkout https://github.com/servo/stylo/commit/1e67ee6a38a4d2e8bd4fbb926d39da07453a0dd9 in a separate directory (recloned the stylo repo)
- Ran DirEqual on the two directories.
- Differences were
   - `.git` folder
   - Modified times for files we different (these seemed to correspond to the time I checkout out the branch)
   - `Cargo.lock` files were present in my original checkout (but these gitignored so not relevant here - note testing upstream branches which don't include our Cargo.lock file)
   - `__pycache__` directory was present in my original checkout (also gitignored)
   
By my reckoning that means that nothing of significance has changed from the repo switchover.

Loirooriol (approved):

PR #187, opened 2025-05-22
Add `ServoRestyleDamage::REBUILD_STACKING_CONTEXT`

Loirooriol:
This restyle damage is used in cases where we need to reconstruct the stacking contexts, because a style change has caused a box to establish or stop establishing a stacking context, but it's unnecessary to rebuild the box tree.

Additionally, `ServoRestyleDamage::REBUILD` is renamed to `REBUILD_BOX`, and old `servo_restyle_damage` values are removed in favor of it.

Servo PR: https://github.com/servo/servo/pull/37088

### DIFF
--- a/style/servo/restyle_damage.rs
+++ b/style/servo/restyle_damage.rs
@@ -5,9 +5,11 @@
 //! The restyle damage is a hint that tells layout which kind of operations may
 //! be needed in presence of incremental style changes.
 
-use crate::computed_values::display::T as Display;
+use crate::computed_values::isolation::T as Isolation;
+use crate::computed_values::mix_blend_mode::T as MixBlendMode;
 use crate::matching::{StyleChange, StyleDifference};
 use crate::properties::ComputedValues;
+use crate::values::computed::basic_shape::ClipPath;
 use std::fmt;
 
 bitflags! {
@@ -17,12 +19,17 @@ bitflags! {
         /// Repaint the node itself.
         ///
         /// Propagates both up and down the flow tree.
-        const REPAINT = 0x01;
+        const REPAINT = 0b001;
+
+        /// Rebuilds the stacking contexts.
+        ///
+        /// Propagates both up and down the flow tree.
+        const REBUILD_STACKING_CONTEXT = 0b011;
 
         /// Any other type of damage, which requires rebuilding all layout objects.
         ///
         /// Propagates both up and down the flow tree.
-        const REBUILD = 0x02;
+        const REBUILD_BOX = 0b111;
     }
 }
 
@@ -46,7 +53,7 @@ impl ServoRestyleDamage {
 
     /// Returns a bitmask indicating that the frame needs to be reconstructed.
     pub fn reconstruct() -> ServoRestyleDamage {
-        ServoRestyleDamage::REBUILD
+        ServoRestyleDamage::REBUILD_BOX
     }
 }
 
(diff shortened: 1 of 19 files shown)


style/servo/restyle_damage.rs
```diff
@@ -89,41 +99,35 @@ fn compute_damage(old: &ComputedValues, new: &ComputedValues) -> ServoRestyleDam
...
+    // one, in order to also force rebuilding stacking contexts.
+    // TODO: `z-index` doesn't always apply, so this restyle damage may be unnecessary.
+    let rebuild_stacking_context_extra = || {
+        old.get_position().z_index.is_auto() != new.get_position().z_index.is_auto() ||
+            (old.get_effects().opacity == 1.0) != (new.get_effects().opacity == 1.0) ||
+            (old.get_effects().mix_blend_mode == MixBlendMode::Normal) !=
+                (new.get_effects().mix_blend_mode == MixBlendMode::Normal) ||
+            (old.get_svg().clip_path == ClipPath::None) !=
+                (new.get_svg().clip_path == ClipPath::None)
+    };
```
nicoburns, line 115: I think the logic here ought to "OR together" every old property, separately "OR together" every new property, and then compare whether the overall result of "establishes a stacking context" has changed. This would allow us to avoid rebuilding stacking contexts in cases where multiple properties change but cancel each other out.

nicoburns: This may also need to be changed in the generated macro?

nicoburns (changes requested):

style/servo/restyle_damage.rs
Loirooriol: This is a good point, but I don't think we can do it like that, because here we don't know if `z-index` is actually applying.

For example, if the style changes from `z-index: 1` to `opacity: 0.5` then both the old and the new styles have non-initial values for properties that can establish a stacking context, but only the new style actually establishes it if `z-index` doesn't apply.

Maybe the style damage should be computed in Servo where we have the real logic for knowing whether a style establishes a stacking context or not.

mrobinson (approved):

style/servo/restyle_damage.rs
Loirooriol: I have updated the comment to clarify this

Loirooriol: Actually, any `z-index` change needs to rebuild the stacking context tree. So it's no longer handled here, and I have applied your suggestion for the remaining properties.

style/servo/restyle_damage.rs
```diff
@@ -89,41 +100,33 @@ fn compute_damage(old: &ComputedValues, new: &ComputedValues) -> ServoRestyleDam
     // This should check every CSS property, as enumerated in the fields of
     // https://doc.servo.org/style/properties/struct.ComputedValues.html
 
+    // Some properties establish a stacking context when they are set to a non-initial value.
+    // In that case, the damage is only set to `ServoRestyleDamage::REPAINT` because we don't
+    // need to rebuild stacking contexts when the style changes between different non-initial
+    // values. This function checks whether any of these properties is set to a value that
+    // guarantees a stacking context, so that we only do the work when this changes.
+    // Note that it's still possible to establish a stackign context when this returns false.
```
mrobinson, line 108:
```suggestion
    // Note that it's still possible to establish a stacking context when this returns false.
```

mrobinson (approved):

nicoburns (approved):

PR #189, opened 2025-05-25
Fix typo

Psychpsyo:
Fixes a typo in a comment.


mrobinson:
This is a downstream port of Stylo which typically only accepts changes necessary for Servo (which later need to be ported upstream). For these sort of changes, please submit them directly upstream to https://github.com/mozilla-firefox/firefox. Thanks!

PR #190, opened 2025-05-28
Have transform and related changes conditionally trigger only overflow damage

mrobinson:
In many cases, changes to transforms do not require rebuilding the box
and laying out again. This change uses RECALCULATE_OVERFLOW restyle damage
in these cases, except when the presence of a transform changes. In those
cases we need to rebuild the box tree.


mrobinson:
We have reworked this to work with overflow damage.

PR #196, opened 2025-06-09
Eliminate `unicode-bidi` dependency

nicoburns:
The `unicode-bidi` crate is only used for it's `Level` type which is just a newtype of `u8`. This PR deletes that function and removes the dependency. Gecko does not appear to use this function.

Corresponding Servo PR (adds the function to Servo): https://github.com/servo/servo/pull/37343


nicoburns:
@Loirooriol That makes sense. This PR now simply deletes the function and removes the dependency. https://github.com/servo/servo/pull/37343 adds the function the `ComputedValuesExt`. The Servo PR should be merged first.

Loirooriol (approved):

PR #197, opened 2025-06-09
Add `AttrValue::ShadowParts`

simonwuelker:
Companion PR for https://github.com/servo/servo/pull/37345
Part of https://github.com/servo/servo/issues/35349

### DIFF
--- a/style/attr.rs
+++ b/style/attr.rs
@@ -6,6 +6,7 @@
 //!
 //! [attr]: https://dom.spec.whatwg.org/#interface-attr
 
+use crate::shadow_parts::ShadowParts;
 use crate::color::{AbsoluteColor, parsing::parse_color_keyword};
 use crate::properties::PropertyDeclarationBlock;
 use crate::shared_lock::Locked;
@@ -72,6 +73,9 @@ pub enum AttrValue {
         String,
         #[ignore_malloc_size_of = "Arc"] Arc<Locked<PropertyDeclarationBlock>>,
     ),
+
+    /// The value of an `exportparts` attribute.
+    ShadowParts(String, ShadowParts),
 }
 
 /// Shared implementation to parse an integer according to
@@ -266,6 +270,11 @@ impl AttrValue {
         AttrValue::Dimension(string, parsed)
     }
 
+    pub fn from_shadow_parts(string: String) -> AttrValue {
+        let shadow_parts = ShadowParts::parse(&string);
+        AttrValue::ShadowParts(string, shadow_parts)
+    }
+
     /// Assumes the `AttrValue` is a `TokenList` and returns its tokens
     ///
     /// ## Panics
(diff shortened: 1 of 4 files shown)


style/attr.rs
```diff
@@ -6,6 +6,7 @@
 //!
 //! [attr]: https://dom.spec.whatwg.org/#interface-attr
 
+use crate::shadow_parts::ShadowParts;
```
jdm, line 9: The module is feature-gated in Servo. Should more of the code in this file be gated as well?

style/shadow_parts.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    // Step 3. Let mappings be an initially empty list of tuples of tokens.
+    // This list will be the result of this algorithm.
+    // NOTE: We return an iterator here - it is up to the caller to turn it into a list
+
+    // Step 4. For each string unparsed mapping in unparsed mappings, run the following substeps:
+    unparsed_mappings.filter_map(|unparsed_mapping| {
+        // Step 4.1 If unparsed mapping is empty or contains only space characters,
+        // continue to the next iteration of the loop.
+        if unparsed_mapping.chars().all(|c| c == ' ') {
```
jdm, line 92: Huh. I did not expect that all() on an empty string returns true, but I just tested it in the playground.

style/shadow_parts.rs
```diff
@@ -0,0 +1,169 @@
...
+            mappings: parse_mapping_list(input).map(|(first,second)| (first.into(), second.into())).collect(),
+        }
+    }
+
+    /// Call the provided callback for each exported part with the given name.
+    pub fn for_each_exported_part<F>(&self, name: &Atom, mut callback: F)
+    where F: FnMut(&AtomIdent) {
+        for (from, to) in &self.mappings {
+            if from == name {
+                println!("Found mapping for {name:?}, it's mapped to {to:?}");
```
jdm: Removed this?

style/shadow_parts.rs
```diff
@@ -0,0 +1,169 @@
...
+                callback(AtomIdent::cast(to));
+            }
+        }
+    }
+
+    pub fn imported_part(&self, name: &Atom) -> Option<&Atom> {
+        println!("get imported mapping for {name:?}");
```
jdm: Remove this?

style/shadow_parts.rs
```diff
@@ -0,0 +1,169 @@
...
+    use super::*;
+
+    #[test]
+    fn parse_valid_mapping() {
+        assert_eq!(parse_part_mapping("foo"), Some(("foo", "foo")), "Single token");
+         assert_eq!(parse_part_mapping("  foo"), Some(("foo", "foo")), "Single token with leading whitespace");
+         assert_eq!(parse_part_mapping("foo "), Some(("foo", "foo")), "Single token with trailing whitespace");
+         assert_eq!(parse_part_mapping("foo:bar"), Some(("foo", "bar")), "Two tokens");
+         assert_eq!(parse_part_mapping("foo:bar "), Some(("foo", "bar")), "Two tokens with trailing whitespace");
+         assert_eq!(parse_part_mapping("foo:bar "), Some(("foo", "bar")), "Two tokens with trailing whitespace");
```
jdm: Duplicate of the previous test case.

style/shadow_parts.rs
simonwuelker: It's a bit more intuitive if you invert the statement twice:
> The closure returns `true` for every character in the empty string

is equivalent to
> There are no characters in the empty string for which the closure returns `false`.

There are no characters for which the closure returns false - because there are no characters at all.

style/attr.rs
simonwuelker: No, because `attr.rs` is itself also feature gated for servo.

https://github.com/servo/stylo/blob/15596275d9bdfcbfc584bdb6618fa2006ac38f29/style/lib.rs#L83-L84

simonwuelker:
Could someone with merge permissions press the button?

PR #198, opened 2025-06-10
Support `type=text` and `type=color` Input Pseudo Element Selector 

stevennovaryo:
Add `::-servo-text-control-inner-editor`, `::-servo-text-control-inner-container`, `::-servo-text-control-placeholder`, and `::-servo-input-color-swatch` pseudo element selector supporting the implementation of of `type=text` and `type=color` input element.

Additionally, preemptively define the standardized version of the selectors `::placeholder` (https://drafts.csswg.org/css-pseudo/#placeholder-pseudo) and `::color-swatch` (https://drafts.csswg.org/css-forms-1/#color-swatch-pseudo). It is done to kept the norm of pseudo element naming consistent (i.e., pseudo name that starts with browser name is browser specific, and mostly private, while others is public and standardized).

These selectors would have a lazy cascade type, although its behavior would differ greatly from a normal lazy pseudo elements. Additionally, implement `PseudoElement::is_color_swatch` for `::-servo-input-color-swatch` which would prevent the tampering of its background color .

Stylo companion of https://github.com/servo/servo/pull/37427 and https://github.com/servo/servo/pull/37483.

Edit: added one more PR that it relates to.

### DIFF
--- a/style/servo/selector_parser.rs
+++ b/style/servo/selector_parser.rs
@@ -71,12 +71,10 @@ pub enum PseudoElement {
     // Private implemented pseudos. Only matchable in UA stylesheet.
     ServoTextControlInnerEditor,
     ServoTextControlInnerContainer,
-    ServoTextControlPlaceholder,
-    ServoInputColorSwatch,
 }
 
 /// The count of all pseudo-elements.
-pub const PSEUDO_COUNT: usize = PseudoElement::ServoInputColorSwatch as usize + 1;
+pub const PSEUDO_COUNT: usize = PseudoElement::ServoTextControlInnerContainer as usize + 1;
 
 impl ToCss for PseudoElement {
     fn to_css<W>(&self, dest: &mut W) -> fmt::Result
@@ -102,8 +100,6 @@ impl ToCss for PseudoElement {
             ColorSwatch => "::color-swatch",
             ServoTextControlInnerEditor => "::-servo-text-control-inner-editor",
             ServoTextControlInnerContainer => "::-servo-text-control-inner-container",
-            ServoTextControlPlaceholder => "::-servo-text-control-placeholder",
-            ServoInputColorSwatch => "::-servo-input-color-swatch",
         })
     }
 }
@@ -195,7 +191,7 @@ impl PseudoElement {
     /// the color swatch inside an `<input>` element.
     #[inline]
     pub fn is_color_swatch(&self) -> bool {
-        matches!(*self, PseudoElement::ColorSwatch | PseudoElement::ServoInputColorSwatch)
+        *self == PseudoElement::ColorSwatch
     }
 
     /// Whether this pseudo-element is eagerly-cascaded.
@@ -247,9 +243,7 @@ impl PseudoElement {
             PseudoElement::Placeholder |
             PseudoElement::ColorSwatch |
             PseudoElement::ServoTextControlInnerEditor |
-            PseudoElement::ServoTextControlInnerContainer |
-            PseudoElement::ServoTextControlPlaceholder |
-            PseudoElement::ServoInputColorSwatch => PseudoElementCascadeType::Lazy,
+            PseudoElement::ServoTextControlInnerContainer => PseudoElementCascadeType::Lazy,
             PseudoElement::DetailsContent |
             PseudoElement::ServoAnonymousBox |
             PseudoElement::ServoAnonymousTable |
@@ -670,8 +664,18 @@ impl<'a, 'i> ::selectors::Parser<'i> for SelectorParser<'a> {
                 }
                 ServoTableWrapper
             },
-            "placeholder" => Placeholder,
-            "color-swatch" => ColorSwatch,
+            "placeholder" => {
+                if !self.in_user_agent_stylesheet() {
+                    return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
+                }
+                Placeholder
+            },
+            "color-swatch" => {
+                if !self.in_user_agent_stylesheet() {
+                    return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
+                }
+                ColorSwatch
+            },
             "-servo-text-control-inner-editor" => {
                 if !self.in_user_agent_stylesheet() {
                     return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
@@ -683,18 +687,6 @@ impl<'a, 'i> ::selectors::Parser<'i> for SelectorParser<'a> {
                     return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
                 }
                 ServoTextControlInnerContainer
-            },
-            "-servo-text-control-placeholder" => {
-                if !self.in_user_agent_stylesheet() {
-                    return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
-                }
-                ServoTextControlPlaceholder
-            },
-            "-servo-input-color-swatch" => {
-                if !self.in_user_agent_stylesheet() {
-                        return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
-                }
-                ServoInputColorSwatch
             },
             _ => return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
 


style/rule_collector.rs
```diff
@@ -104,12 +104,6 @@ where
 
         let matches_user_and_content_rules = rule_hash_target.matches_user_and_content_rules();
 
-        // Gecko definitely has pseudo-elements with style attributes, like
-        // ::-moz-color-swatch.
-        debug_assert!(
-            cfg!(feature = "gecko") || style_attribute.is_none() || pseudo_element.is_none(),
-            "Style attributes do not apply to pseudo-elements"
-        );
```
stevennovaryo, line 112: We are also allowing style attributes for pseudo element. Notice that, in Servo, this will applies only for the pseudo element with a concrete element behind it (i.e. implemented pseudo element).

stevennovaryo:
Could you help with the review of the companion PR as well? Thank you! @xiaochengh

style/servo/selector_parser.rs
```diff
@@ -190,7 +202,7 @@ impl PseudoElement {
     /// Whether this pseudo-element is the ::-moz-color-swatch pseudo.
```
xiaochengh:
```suggestion
    /// Whether this pseudo-element is the ::-servo-input-color-swatch pseudo.
```

stevennovaryo: Done thanks!

style/servo/selector_parser.rs
```diff
@@ -238,7 +250,11 @@ impl PseudoElement {
             },
             PseudoElement::Backdrop |
             PseudoElement::DetailsSummary |
-            PseudoElement::Marker  => PseudoElementCascadeType::Lazy,
+            PseudoElement::Marker |
+            PseudoElement::ServoTextControlInnerEditor |
+            PseudoElement::ServoTextControlInnerContainer |
+            PseudoElement::ServoTextControlPlaceholder |
+            PseudoElement::ServoInputColorSwatch => PseudoElementCascadeType::Lazy,
```
nicoburns: As these are standardised pseudo elements from a CSS spec, I don't think they should be `-servo` prefixed.

style/servo/selector_parser.rs
```diff
@@ -708,6 +724,30 @@ impl<'a, 'i> ::selectors::Parser<'i> for SelectorParser<'a> {
...
+                }
+                ServoTextControlInnerEditor
+            },
+            "-servo-text-control-inner-container" => {
+                if !self.in_user_agent_stylesheet() {
+                    return Err(location.new_custom_error(SelectorParseErrorKind::UnexpectedIdent(name.clone())))
+                }
+                ServoTextControlInnerContainer
+            },
+            "-servo-text-control-placeholder" => {
```
nicoburns: Is the plan to eventually expose these to author stylesheets?

stevennovaryo: Sorry I was already offline yesterday. Yes, that is the plan, it will be exposed using its standardized version.

Ideally if Servo have a proper UA Shadow DOM (or equivalent structure) for `<input>` in Servo, the implementation could be more simple. But we are yet to do that, and I afraid it would bloat the implementation, so we are doing this step by step. But it does make sense to add the definition of the standardized version as well.

style/servo/selector_parser.rs
stevennovaryo: These private pseudo selectors for UA stylesheet, I have defined `::placeholder` and `::color-swatch` as well.

nicoburns: Why does the UA stylesheet need a separate selector? Can't it just use the standard one?

style/servo/selector_parser.rs
nicoburns: I think it would make sense to use the standardized version now, and just hide it from author stylesheets until we're ready to expose it (unless there's some reason why this is a bad idea?)

style/servo/selector_parser.rs
```diff
@@ -86,6 +98,12 @@ impl ToCss for PseudoElement {
             ServoAnonymousTableRow => "::-servo-anonymous-table-row",
             ServoTableGrid => "::-servo-table-grid",
             ServoTableWrapper => "::-servo-table-wrapper",
+            Placeholder => "::placeholder",
+            ColorSwatch => "::color-swatch",
+            ServoTextControlInnerEditor => "::-servo-text-control-inner-editor",
+            ServoTextControlInnerContainer => "::-servo-text-control-inner-container",
+            ServoTextControlPlaceholder => "::-servo-text-control-placeholder",
+            ServoInputColorSwatch => "::-servo-input-color-swatch",
```
mrobinson: Instead of exposing a pseudo for each of component of a control, couldn't we simply match against them using CSS selectors? I'm a bit confused about why this is necessary.

stevennovaryo: Ideally, we need a matching that would:
- Does not mess the cascading. (hence inline styles not working)
- Does not results in reparsing of styles and lightweight. (hence bare inline styles and embedding stylesheet being not optimal).
- Private, so it would not be disturb or be disturbed by user styles. (hence public selector in UA stylesheet not working)

@xiaochengh also mentioned this in the following comment. https://github.com/servo/servo/pull/36994#issuecomment-2879323823

If we have a lazily initiated stylesheet, which is read-only, and can be safely shared between UA shadow DOM element, it would pass those requirements. Albeit, with a bit heavier UA shadow DOM. Though, that should not should block the implementations of Shadow DOM element.

stevennovaryo: Practically, this is no different than the plan to expose `::-moz-progress-bar` (or `::-servo-progress-bar`). https://github.com/servo/servo/pull/35531#discussion_r1969290143

mrobinson: So I'm imagining a single pseudo-element like `::servo-form-control` and then classes and selectors would be used to select individual parts inside that form control.

stevennovaryo: Ah, I get what you mean now, I suppose that would make more sense, to avoid bloating `PseudoElement` as well.

stevennovaryo: I did a little bit of digging for the idea. I suppose Stylo does not support further chaining (probably something similar to https://bugzilla.mozilla.org/show_bug.cgi?id=1954142#c2). So, in other words, the `input::-servo-form-control#-servo-text-control-inner-container` would not work, and we need an additional container above it, which is not ideal. And, if we are using traits that could be matched publicly, it could technically match `::input #-servo-text-control-inner-container`, which is not desired as well.

I noticed of opportunity having the selector to behave like `:host` selector which would match the shadow host instead. But this would require much more further works.

I believe it would make sense to go back to how do other browser do this first. I.e., defining an internal pseudo element for each elements that should be matched since the affect is quite minuscule (i.e., we only have one more selector). And then, we could do more in depth analysis and improve it later.

style/servo/selector_parser.rs
stevennovaryo: After having more thoughts about this, It does make sense. I am thinking about reducing the dependency of changes later. But it would not make sense that we are exposing a false positive (i.e., passing the WPT for parsing `::placeholder` but are not actually implementing it).

style/servo/selector_parser.rs
stevennovaryo: This is used to handle the elements with "UA shadow DOM" (or anonymous tree in Firefox). It basically refer to elements that actually have some DOM element(s) inside it (e.g. `<input>` having inner container with text, or buttons). 

Major browsers are handling this with the internal pseudo elements that is usable only inside UA stylesheet (e.g., Firefox's `::-moz-text-control-editing-root` https://searchfox.org/mozilla-central/source/layout/style/res/forms.css#135).

mrobinson (changes requested):
I'm still unclear why this change includes this particular collection of pseudo-element selectors. Typically we add support for a pseudo as we add them in Servo. I think it might make sense to do this bit-by-bit along with the Servo changes that go with them.

stevennovaryo:
> I'm still unclear why this change includes this particular collection of pseudo-element selectors. Typically we add support for a pseudo as we add them in Servo. I think it might make sense to do this bit-by-bit along with the Servo changes that go with them.

That is exactly what the PRs (one in Stylo and two others in Servo) are doing right now.

I know that we are splitting the Servo PRs apart, but essentially, the other PR is a reland of a merged PR, with minimized changes, so it could be merged quick after the PRs that it depends on do so.

style/servo/selector_parser.rs
xiaochengh: > Why does the UA stylesheet need a separate selector? Can't it just use the standard one?

AFAICT this is the standard approach in Blink, WebKit and Gecko.

Typically a UA shadow DOM contains more elements than the number of web-exposed pseudo elements. Therefore we are introducing private selectors for them.

As background, we tried to not involve private selectors but adding a stylesheet directly into the UA shadow DOM. It causes a major performance regression, though, and got reverted.

style/servo/selector_parser.rs
xiaochengh: > I think it would make sense to use the standardized version now, and just hide it from author stylesheets until we're ready to expose it (unless there's some reason why this is a bad idea?)

This makes sense. No need to introduce a private selector if there's already a standard one.

> After having more thoughts about this, It does make sense. I am thinking about reducing the dependency of changes later. But it would not make sense that we are exposing a false positive (i.e., passing the WPT for parsing ::placeholder but are not actually implementing it).

As long as we hide it from author stylesheets for now we won't introduce such false positives.

xiaochengh:
@mrobinson Could you elaborate?

The Servo-side change is at https://github.com/servo/servo/pull/37427, which adds these element into Servo. The patch is small and probably doesn't need further breaking down. And we just need this one to land first.

stevennovaryo:
Closed because this PR was split into https://github.com/servo/stylo/pull/208 and https://github.com/servo/stylo/pull/209.

PR #195, opened 2025-06-06
Implement `::part` selector support

simonwuelker:
Companion PR for https://github.com/servo/servo/pull/37307

PR #199, opened 2025-06-13
Pin exact version of web_atoms

simonwuelker:
We depend on `0.1.3`, `0.1.x` is not good enough.
See https://github.com/servo/stylo/pull/195#issuecomment-2970599602.


nicoburns:
I would note that this doesn't pin an exact version of web_atoms. It will still allow `1.3.4`, etc if/when those get published.

mrobinson (approved):

PR #201, opened 2025-06-14
Use `Cell` instead of `RefCell` for storing `ThreadState`

simonwuelker:
I profiled servo loading a local copy of https://html.spec.whatwg.org and saw around 15% of the time being spent in `style::thread_state::get()`. Most calls are inside assertions, so they likely go away in a production build, but that's still way too much time. Arguably we shouldn't assert on the thread state that much in the first place, especially given that layout and script are now in the same thread, but replacing the `RefCell` with a `Cell` already takes the time down to 10%.

Additionally, this change renames all one-letter variables for clarity and removes the `thread_types` macro because it is unnecessary (Removing it *decreases* the LoC!)

I wasn't sure whether I should submit the patch here or upstream, but gecko doesn't use the thread state so i figured this repo was the right place.

### DIFF
--- a/style/thread_state.rs
+++ b/style/thread_state.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-//! Supports dynamic assertions in about what sort of thread is running and
+//! Supports dynamic assertions about what sort of thread is running and
 //! what state it's in.
 
 #![deny(missing_docs)]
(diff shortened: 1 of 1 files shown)


jdm:
>Arguably we shouldn't assert on the thread state that much in the first place, 

This isn't true for all of the rayon worker threads for parallel style recalculation and layout.

style/thread_state.rs
```diff
@@ -27,38 +27,34 @@ bitflags! {
...
+        self.contains(ThreadState::IN_WORKER)
+    }
 
-        $(
-            #[allow(missing_docs)]
-            pub fn $fun(self) -> bool {
-                self.contains($flag)
-            }
-        )*
+    #[allow(missing_docs)]
```
nicoburns: Can we get a doc comment like the one for `is_worker` for `is_script` and `is_layout`?

nicoburns (approved):
This looks equivalent to the original version to me. Very nice to get rid of the macro!

simonwuelker:
This change is approved and ready to be merged as far as I'm concerned. Does someone with the necessary permissions want to push the button?

PR #208, opened 2025-06-26
Define `::color-swatch` Pseudo Element of Input `type=color` 

stevennovaryo:
Define `::color-swatch` pseudo element https://drafts.csswg.org/css-forms-1/#color-swatch-pseudo. It represents the element that display the chosen color value of Input `type=color`. Particularly, `::color-swatch` is a pseudo element with a concrete DOM element behind it, and it is defined as lazy pseudo element. 

Stylo's companion of https://github.com/servo/servo/pull/37427

PR #209, opened 2025-06-27
Define Implemented Pseudo Element for Input `type=text`

stevennovaryo:
Undirectly depends on: https://github.com/servo/stylo/pull/208 

Define `::placeholder`, `::-servo-text-control-inner-editor`, and `::-servo-text-control-inner-container`. The implemented pseudo elements that support style matching for Input `type=text` UA Shadow DOM. All of these pseudos will be defined as a lazy and accessible only in UA stylesheet (private). 

Contrary to `::placeholder` status as a standardized pseudo element, it will be defined as a private pseudo for now due to the lack of property restriction and placeholder for `<textarea>` being not implemented.

Stylo side of https://github.com/servo/servo/pull/37483


Loirooriol:

PR #212, opened 2025-07-02
Add support for the ::color-swatch pseudo-element

stevennovaryo:
Define `::color-swatch` pseudo element https://drafts.csswg.org/css-forms-1/#color-swatch-pseudo. It represents the element that display the chosen color value of Input `type=color`. Particularly, `::color-swatch` is a pseudo element with a concrete DOM element behind it, and it is defined as lazy pseudo element. 

The pseudo element would be public because of no additional requirement specified (e.g. property restriction) and the element is defined in Servo.

Stylo's companion of https://github.com/servo/servo/pull/37427

This is a reopened PR of https://github.com/servo/stylo/pull/208 due to the new upstream.

### DIFF
--- a/style/servo/selector_parser.rs
+++ b/style/servo/selector_parser.rs
@@ -55,6 +55,13 @@ pub enum PseudoElement {
     DetailsSummary,
     DetailsContent,
     Marker,
+
+    // Implemented pseudos. These pseudo elements are representing the
+    // elements within an UA shadow DOM, and matching the elements with
+    // their appropriate styles.
+    ColorSwatch,
+
+    // Servo-specific pseudos
     ServoAnonymousBox,
     ServoAnonymousTable,
     ServoAnonymousTableCell,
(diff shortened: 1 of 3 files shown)


style/servo/selector_parser.rs
```diff
@@ -173,10 +181,11 @@ impl PseudoElement {
         false
     }
 
-    /// Whether this pseudo-element is the ::-moz-color-swatch pseudo.
+    /// Whether this pseudo-element is the pseudo element representing
+    /// the color swatch inside an `<input>` element.
```
Loirooriol: Seems a bit weird to repeat "pseudo-element", especially since one has an hyphen and the order doesn't.

```suggestion
    /// Whether this pseudo-element is representing the color swatch
    /// inside an `<input>` element.
```

style/servo/selector_parser.rs
```diff
@@ -55,6 +55,13 @@ pub enum PseudoElement {
     DetailsSummary,
     DetailsContent,
     Marker,
+
+    // Implemented pseudos. These pseudo elements are representing the
+    // elements within an UA shadow DOM, and matching the elements with
+    // their appropriate styles.
+    ColorSwatch,
+
+    // Servo specific pseudos
```
Loirooriol:
```suggestion
    // Servo-specific pseudos
```

Loirooriol (approved):
Looks good with nits

Loirooriol:
This comment seems wrong now:

https://github.com/servo/stylo/blob/ece49719ac02599750bd15901ecd7be24e96a26a/style/dom.rs#L673

style/servo/selector_parser.rs
stevennovaryo: The new comments looks more neat, thanks!

PR #213, opened 2025-07-03
Animation.rs: Iterate declared transition backwards

yezhizhen:
Implement `DoubleEndedIterator` for `TransitionPropertyIterator` to iterate declared transition backwards:
[Spec](https://www.w3.org/TR/css-transitions-1/#transitions): "If a property is specified multiple times in the value of transition-property (either on its own, via a shorthand that contains it, or via the all value), then the transition that starts uses the duration, delay, and timing function at the index corresponding to the last item in the value of transition-property that calls for animating that property. "

Addresses Servo issue: [servo #37164](https://github.com/servo/servo/issues/37164)
Servo PR: https://github.com/servo/servo/pull/37176

Rebase of #192


mrobinson:
@yezhizhen Thanks for posting this. Since this PR is much bigger than I expected and might take a while to review and ensure we can get it upstream, I've posted your original version as well here: #214. I didn't realize that it would be so much work to add reverse iteration support here, sorry. I don't want to block your work any longer than I already have, so my hope is we can land the less optimal version first and then land something like this later.

PR #215, opened 2025-07-03
Vendor markupsafe; Add macOS and Windows CI jobs

nicoburns:
# Objective

- Fix https://github.com/servo/stylo/issues/218
- Add CI checks to ensure that Stylo works (or at least compiles) on all target platforms

## Changes made
- Vendor `markupsafe` as python source code
- Configure python not to convert newlines when writing the generated code (fixes windows build)
- Add CI jobs for macOS and Windows

## Context

I'm seeing Blitz [CI failures](https://github.com/DioxusLabs/blitz/actions/runs/16048942377/job/45288450123) on macOS with Stylo 0.5. ~~It works locally so I'm not quite sure what the issue is yet, but it seems be related to https://github.com/servo/stylo/commit/a3ef6a2749b2c28a32c62cd9b8954950507dc437~~ EDIT: see https://github.com/servo/stylo/issues/218


style/properties/vendored_python/markupsafe/__init__.py
Loirooriol, line 1: Should this include markupsafe's license to comply with

> 1.  Redistributions of source code must retain the above copyright
>    notice, this list of conditions and the following disclaimer.

?

nicoburns: Good point. I've added the LICENSE.txt file for markupsafe.

style/properties/build.py
```diff
@@ -114,7 +115,7 @@ def write(directory, filename, content):
     if not os.path.exists(directory):
         os.makedirs(directory)
     full_path = os.path.join(directory, filename)
-    open(full_path, "w", encoding="utf-8").write(content)
+    open(full_path, "w", encoding="utf-8", newline="").write(content)
```
Loirooriol, line 118: Why is this needed?

delan (approved):

style/properties/build.py
Loirooriol: OK, it's in the description:

> Configure python not to convert newlines when writing the generated code (fixes windows build)

nicoburns: Without it I was getting "bare CR not allowed in doc-comment" on windows. So I think Python was converting the line endings to DOS line endings when it wrote the file (`""` value for `newline` means "don't modify" - https://docs.python.org/3/library/functions.html#open)

Loirooriol:
Please don't forget to bump Servo when you land changes into Stylo.
I don't expect any problem, but sometimes things happen and then the next person who wants to land something needs to deal with it.
https://github.com/servo/servo/pull/38351

nicoburns:
> Please don't forget to bump Servo when you land changes into Stylo. I don't expect any problem, but sometimes things happen and then the next person who wants to land something needs to deal with it. [servo/servo#38351](https://github.com/servo/servo/pull/38351)

Good point! I'll try to make sure I remember this in future.

PR #221, opened 2025-08-03
Drop obsolete `layout.css.transition-behavior.enabled` flag

Loirooriol:
There is no point in setting this flag to true when it's no longer doing anything, `transition-behavior` is now enabled unconditionally.


Loirooriol:
Servo PR: https://github.com/servo/servo/pull/38449

PR #222, opened 2025-08-04
Simply restyle damage macros (and convert them to be plain functions)

nicoburns:
Changes made:
- Convert `restyle_damage_*` macros to plain functions.
- Rather than passing the `damage` object, a `RestyleDamage` flag to apply, and an optional extra condition, we make the return a boolean and handle the rest in regular code.
- Use `if-else` instead of `||` for short-circuiting.

Servo PR https://github.com/servo/servo/pull/38465

### DIFF
--- a/style/properties/properties.mako.rs
+++ b/style/properties/properties.mako.rs
@@ -2964,28 +2964,15 @@ const_assert!(std::mem::size_of::<longhands::${longhand.ident}::SpecifiedValue>(
 
 % if engine == "servo":
 % for effect_name in ["repaint", "recalculate_overflow", "rebuild_stacking_context", "rebuild_box"]:
-    macro_rules! restyle_damage_${effect_name} {
-        ($old: ident, $new: ident, $damage: ident, [ $($effect:expr),* ]) => ({
-            restyle_damage_${effect_name}!($old, $new, $damage, [$($effect),*], false)
-        });
-        ($old: ident, $new: ident, $damage: ident, [ $($effect:expr),* ], $extra:expr) => ({
-            if
-                % for style_struct in data.active_style_structs():
-                    % for longhand in style_struct.longhands:
-                        % if effect_name in longhand.servo_restyle_damage.split() and not longhand.logical:
-                            $old.get_${style_struct.name_lower}().${longhand.ident} !=
-                            $new.get_${style_struct.name_lower}().${longhand.ident} ||
-                        % endif
-                    % endfor
-                % endfor
-
-                $extra || false {
-                    $damage.insert($($effect)|*);
-                    true
-            } else {
-                false
-            }
-        });
-    }
+pub(crate) fn restyle_damage_${effect_name} (old: &ComputedValues, new: &ComputedValues) -> bool {
+    % for style_struct in data.active_style_structs():
+        % for longhand in style_struct.longhands:
+            % if effect_name in longhand.servo_restyle_damage.split() and not longhand.logical:
+                old.get_${style_struct.name_lower}().${longhand.ident} != new.get_${style_struct.name_lower}().${longhand.ident} ||
+            % endif
+        % endfor
+    % endfor
+    false
+}
 % endfor
 % endif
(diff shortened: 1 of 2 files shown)


style/servo/restyle_damage.rs
```diff
@@ -146,6 +147,13 @@ fn compute_damage(old: &ComputedValues, new: &ComputedValues) -> ServoRestyleDam
             old.get_effects().filter.0.is_empty() != new.get_effects().filter.0.is_empty()
     };
 
+    restyle_damage_rebuild_box(old, new) && rebuild_box_extra()
```
Loirooriol: Doesn't this need to be OR?

```suggestion
    restyle_damage_rebuild_box(old, new) || rebuild_box_extra()
```

And at this point you can inline `rebuild_box_extra()`

Loirooriol (changes requested):

style/servo/restyle_damage.rs
nicoburns: Ah yes, well spotted.

style/servo/restyle_damage.rs
```diff
@@ -159,25 +165,24 @@ fn compute_damage(old: &ComputedValues, new: &ComputedValues) -> ServoRestyleDam
...
-        [ServoRestyleDamage::RECALCULATE_OVERFLOW]
-    ) || restyle_damage_rebuild_stacking_context!(
-        old,
-        new,
-        damage,
-        [ServoRestyleDamage::REBUILD_STACKING_CONTEXT],
-        guarantees_stacking_context(old) != guarantees_stacking_context(new)
-    ) || restyle_damage_repaint!(old, new, damage, [ServoRestyleDamage::REPAINT]);
+    restyle_damage_rebuild_stacking_context(old, new)
+        && guarantees_stacking_context(old) != guarantees_stacking_context(new)
```
Loirooriol: Also `||`

nicoburns: Fixed

style/servo/restyle_damage.rs
```diff
@@ -186,3 +166,30 @@ fn compute_damage(old: &ComputedValues, new: &ComputedValues) -> ServoRestyleDam
...
+    /// need to rebuild stacking contexts when the style changes between different non-initial
+    /// values. This function checks whether any of these properties is set to a value that
+    /// guarantees a stacking context, so that we only do the work when this changes.
+    /// Note that it's still possible to establish a stacking context when this returns false.
+    pub fn guarantees_stacking_context(&self) -> bool {
+        self.get_effects().opacity != 1.0
+            || self.get_effects().mix_blend_mode != MixBlendMode::Normal
+            || self.get_svg().clip_path != ClipPath::None
+            || self.get_box().isolation == Isolation::Isolate
+    }
```
nicoburns, line 181: @Loirooriol For the `mix_blend_mode` and `clip_path` checks, the version of this function on `main`: https://github.com/servo/stylo/blob/0d8dd5455daf3bc445312032a0a430e11a817390/style/servo/restyle_damage.rs#L155-L160 always checks `old.` (via closure capture) even when `new` is passed as the closure argument.

That looks wrong to me and I've changed it here, but could you confirm that this was indeed a mistake and not intentional.

Loirooriol: Mm yeah, seems it was wrong.

Loirooriol (approved):

style/properties/properties.mako.rs
```diff
@@ -2964,28 +2964,15 @@ const_assert!(std::mem::size_of::<longhands::${longhand.ident}::SpecifiedValue>(
...
-                $extra || false {
-                    $damage.insert($($effect)|*);
-                    true
-            } else {
-                false
-            }
-        });
-    }
+pub(crate) fn restyle_damage_${effect_name} (old: &ComputedValues, new: &ComputedValues) -> bool {
+    % for style_struct in data.active_style_structs():
```
emilio, line 2968: FWIW you really want to check `arc::ptr_eq(old.get_${style_struct}(), new${style_struct}())` or so to avoid most of these comparisons. That's equivalent to what gecko does in `CalcStyleDifference`: https://searchfox.org/mozilla-central/rev/f6a806c38c459e0e0d797d264ca0e8ad46005105/layout/style/ComputedStyle.cpp#133

PR #225, opened 2025-08-07
RestyleDamage: Ensure correctness with externally set bits

nicoburns:
Ensure's that methods defined by the `bitflags` crate (`all`, `contains`, `empty`, etc) take into account the possibility of externally set bits. See: https://docs.rs/bitflags/latest/bitflags/#externally-defined-flags


Loirooriol (commented):
IIRC in some places we were using `truncate()` or such to remove external bits. Seems your patch would break that.

Issue #226, opened 2025-08-07
Optimize style comparison when computing restyle damage


Loirooriol:
https://github.com/servo/stylo/pull/222#discussion_r2259969103

> FWIW you really want to check `arc::ptr_eq(old.get_${style_struct}(), new${style_struct}())` or so to avoid most of these comparisons. That's equivalent to what gecko does in `CalcStyleDifference`: https://searchfox.org/mozilla-central/rev/f6a806c38c459e0e0d797d264ca0e8ad46005105/layout/style/ComputedStyle.cpp#133

PR #227, opened 2025-08-07
Optimise style comparison when computing damage

nicoburns:
Fixes https://github.com/servo/stylo/issues/226

This compares the pointers of every single style_struct, which may not be necessary. We may instead wish to inject style_struct pointer comparison into each generated comparison. In either case it probably makes sense to compare the top-level pointer to the entire `ComputedValues`, and bail early if they are equal. That captures the (common) case where the styles for a node are entirely unchanged.

### DIFF
--- a/style/properties/properties.mako.rs
+++ b/style/properties/properties.mako.rs
@@ -1788,6 +1788,34 @@ impl ComputedValues {
       self.custom_properties() == other.custom_properties()
     }
 
+    /// Runs cheap ptr_eq checks to determine whether the styles may differ:
+    ///
+    /// - If this function returns false, then all styles are definitely equal
+    /// - If this function returns true, then styles may or may not be equal
+    ///   and further checks (e.g. full deep equals) is necessary.
+    ///
+    /// Checks by:
+    ///   - Doing a pointer comparison on the top-level pointer. If the pointers
+    ///     point to the exact same struct then they must be equal.
+    ///   - Doing a pointer comparison on each "style struct"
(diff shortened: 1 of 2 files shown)


style/properties/properties.mako.rs
```diff
@@ -3004,11 +3004,26 @@ const_assert!(std::mem::size_of::<longhands::${longhand.ident}::SpecifiedValue>(
 % for effect_name in ["repaint", "recalculate_overflow", "rebuild_stacking_context", "rebuild_box"]:
 pub(crate) fn restyle_damage_${effect_name} (old: &ComputedValues, new: &ComputedValues) -> bool {
     % for style_struct in data.active_style_structs():
+        <% longhands_affected = [effect_name in longhand.servo_restyle_damage.split() for longhand in style_struct.longhands if not longhand.logical] %>
+        % if any(longhands_affected):
+        let old_${style_struct.name_lower} = old.get_${style_struct.name_lower}();
+        let new_${style_struct.name_lower} = new.get_${style_struct.name_lower}();
+        if !std::ptr::eq(old_${style_struct.name_lower}, new_${style_struct.name_lower}) {
+        % if all(longhands_affected):
+            return true;
```
Loirooriol: Could it happen that the pointers are different but the values are the same?

nicoburns: Hmm... This is a good point. I may remove this optimisation for now and we can add it back later if desired.

nicoburns: Removed

PR #232, opened 2025-08-19
Format Stylo code

nicoburns:
The same formatting has landed upstream. So in theory if we apply it here and then squash that should make the next sync much smoother. This could also be applied as part of the sync process.


jdm (approved):

Loirooriol:
> that should make the next sync much smoother

Actually the opposite :sweat_smile:

nicoburns:
Oh no. Did you squash the Servo changes before attempting a rebase?

PR #238, opened 2025-09-08
Replace `fxhash`, which is unmaintained, with `rustc-hash`.

orium:
See https://rustsec.org/advisories/RUSTSEC-2025-0057.html.


Loirooriol:
We prefer to not diverge from Firefox unless necessary. Can you send your changes there?
https://bugzilla.mozilla.org/show_bug.cgi?id=1987258

PR #243, opened 2025-09-25
Add support for `-webkit-fill-available` sizing keyword

Loirooriol:
This keyword is de-facto required for web compatibility, Firefox is also enabling it (https://bugzil.la/1988938).

Servo will just alias it to `stretch`.

Servo PR: https://github.com/servo/servo/pull/39492

### DIFF
--- a/style/values/generics/length.rs
+++ b/style/values/generics/length.rs
@@ -164,7 +164,6 @@ pub enum GenericSize<LengthPercent> {
     #[cfg(feature = "gecko")]
     #[animation(error)]
     MozAvailable,
-    #[cfg(feature = "gecko")]
     #[animation(error)]
     WebkitFillAvailable,
     #[animation(error)]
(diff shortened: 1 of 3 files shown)


yezhizhen (dismissed):

style/values/generics/length.rs
```diff
@@ -260,14 +256,12 @@ where
 {
     fn collect_completion_keywords(f: style_traits::KeywordsCollectFn) {
         LP::collect_completion_keywords(f);
-        f(&["none", "stretch", "fit-content"]);
+        f(&["none", "stretch", "fit-content", "max-content", "min-content"]);
```
yezhizhen, line 259: Does this also enable max-content & min-content for Servo?

I didn't see it in the PR description.

Loirooriol: Maybe I should have added a comment. `min-content` and `max-content` are already enabled in Servo. This `collect_completion_keywords` is basically for autocomplete in devtools, and I think it's not even used from Servo.

But while I was at it, I fixed the wrong things:
 - `min-content`, `max-content` and now `-webkit-fill-available` aren't Gecko-only
 - `-webkit-fill-available` may be disabled behind a pref

PR #248, opened 2025-10-04
Make servo feature default for `stylo_traits`

nicoburns:
We are already doing this (and maintaining the patch) for the `stylo` crate. This wasn't previously necessary for `stylo_traits`, but that crate now also requires one of (servo, gecko) in order to compile.


Loirooriol:
It's fine either way, just an idea

PR #249, opened 2025-10-04
Use workspace versions

nicoburns:
These changes are aimed at making version bumps slightly easier as the version numbers will mostly all be in one place.

Changes made:
- The "main" version (currently 0.8.0) is defined in the workspace `Cargo.toml` and inherited by each package that uses it.
- The dependencies between crates within this repo are defined at the workspace level and inherited by each package.

This means that the process to bump the "main" version would just be to replace all the instances of that version number in the workspace `Cargo.toml`.

For the crates which don't follow the "main" version, you would need to update the version in that crate's `Cargo.toml` and the `workspace.dependencies` section of the workspace `Cargo.toml` (which should hopefully be better than having to update each crate that depends on the crate with the updated version).


Cargo.toml
```diff
@@ -16,3 +16,23 @@ members = [
...
+
+# in-repo dependencies (main version)
+malloc_size_of  = { version = "0.8.0",  path = "./malloc_size_of", package = "stylo_malloc_size_of", features = ["servo"] }
+static_prefs    = { version = "0.8.0",  path = "./stylo_static_prefs", package = "stylo_static_prefs" }
+style_config    = { version = "0.8.0",  path = "./stylo_config", package = "stylo_config"}
+stylo_atoms     = { version = "0.8.0",  path = "./stylo_atoms" }
+dom             = { version = "0.8.0",  path = "./stylo_dom", package = "stylo_dom" }
+style_traits    = { version = "0.8.0",  path = "./style_traits", features = ["servo"], package = "stylo_traits"}
+style_derive    = { version = "0.8.0",  path = "./style_derive", package = "stylo_derive"}
+stylo           = { version = "0.8.0",  path = "./style" }
```
Loirooriol: Nit: missing newline at the end of the file

Loirooriol (approved):
Nice! It wasn't much convenient beforehand

Issue #235, opened 2025-08-20
Defaulting to `rebuild_box` is bad for performance


Loirooriol:
From Emilio's review in https://phabricator.services.mozilla.com/D261578#inline-1434184

> This is a bad default for performance, maybe you should require it?

https://github.com/servo/stylo/blob/8d42ac8daefef66c677ad90cbc8066362340d9a5/style/properties/data.py#L378

mrobinson:
I'm happy to require it if it does not add a big maintenance burden.

PR #252, opened 2025-10-10
Mark some text decoration properties as overflow/repaint damage

nicoburns:
Marks some text properties which I believe to affect paint/overflow only as such.

It is very common for `:hover` styles for inline spans (e.g. links) to update the `text-decoration` properties without updating any layout-affecting properties, and these updates can be triggered by events like "mouse move" and "scroll", so it is a nice optimisation to avoid layout / box construction for these updates.

Helps with https://github.com/servo/stylo/issues/235

Servo PR: https://github.com/servo/servo/pull/39769

### DIFF
--- a/style/properties/longhands/inherited_text.mako.rs
+++ b/style/properties/longhands/inherited_text.mako.rs
@@ -176,6 +176,7 @@ ${helpers.predefined_type(
     ignored_when_colors_disabled=True,
     simple_vector_bindings=True,
     spec="https://drafts.csswg.org/css-text-decor-3/#text-shadow-property",
+    servo_restyle_damage="repaint",
     affects="overflow",
 )}
 
--- a/style/properties/longhands/text.mako.rs
+++ b/style/properties/longhands/text.mako.rs
@@ -36,7 +36,7 @@ ${helpers.predefined_type(
     initial_specified_value="specified::TextDecorationLine::none()",
     animation_type="discrete",
     spec="https://drafts.csswg.org/css-text-decor/#propdef-text-decoration-line",
-    servo_restyle_damage="rebuild_box",
+    servo_restyle_damage="recalculate_overflow",
     affects="overflow",
 )}
 
@@ -47,6 +47,7 @@ ${helpers.single_keyword(
     gecko_enum_prefix="StyleTextDecorationStyle",
     animation_type="discrete",
     spec="https://drafts.csswg.org/css-text-decor/#propdef-text-decoration-style",
+    servo_restyle_damage="recalculate_overflow",
     affects="overflow",
 )}
 
@@ -70,6 +71,7 @@ ${helpers.predefined_type(
     initial_specified_value="specified::Color::currentcolor()",
     ignored_when_colors_disabled=True,
     spec="https://drafts.csswg.org/css-text-decor/#propdef-text-decoration-color",
+    servo_restyle_damage="recalculate_overflow",
     affects="paint",
 )}
 


style/properties/longhands/inherited_text.mako.rs
```diff
@@ -333,6 +334,7 @@ ${helpers.predefined_type(
     "computed::LengthPercentageOrAuto::auto()",
     engines="gecko",
     spec="https://drafts.csswg.org/css-text-decor-4/#underline-offset",
+    servo_restyle_damage="recalculate_overflow",
```
Loirooriol: No need to add this since it's gecko-only (`engines="gecko"`)

style/properties/longhands/inherited_text.mako.rs
```diff
@@ -344,6 +346,7 @@ ${helpers.predefined_type(
     engines="gecko",
     animation_type="discrete",
     spec="https://drafts.csswg.org/css-text-decor-3/#text-underline-position-property",
+    servo_restyle_damage="recalculate_overflow",
```
Loirooriol: Ditto

style/properties/longhands/inherited_text.mako.rs
```diff
@@ -355,6 +358,7 @@ ${helpers.predefined_type(
     engines="gecko",
     animation_type="discrete",
     spec="https://drafts.csswg.org/css-text-decor-4/#text-decoration-skip-ink-property",
+    servo_restyle_damage="recalculate_overflow",
```
Loirooriol: Ditto

style/properties/longhands/text.mako.rs
```diff
@@ -59,6 +60,7 @@ ${helpers.predefined_type(
     boxed=True,
     gecko_pref="layout.css.text-decoration-trim.enabled",
     spec="https://drafts.csswg.org/css-text-decor-4/#text-decoration-skip-inset-property",
+    servo_restyle_damage="recalculate_overflow",
```
Loirooriol: Ditto

style/properties/longhands/text.mako.rs
```diff
@@ -92,5 +95,6 @@ ${helpers.predefined_type(
     engines="gecko",
     initial_specified_value="generics::text::GenericTextDecorationLength::Auto",
     spec="https://drafts.csswg.org/css-text-decor-4/#text-decoration-width-property",
+    servo_restyle_damage="recalculate_overflow",
```
Loirooriol: Ditto

style/properties/longhands/inherited_text.mako.rs
nicoburns: I think it would make sense to define this for all standard properties. But I'll remove for now.

style/properties/longhands/text.mako.rs
```diff
@@ -70,6 +71,7 @@ ${helpers.predefined_type(
     initial_specified_value="specified::Color::currentcolor()",
     ignored_when_colors_disabled=True,
     spec="https://drafts.csswg.org/css-text-decor/#propdef-text-decoration-color",
+    servo_restyle_damage="repaint",
```
Loirooriol: This property is used while building the stacking context tree. Not sure if `repaint` is safe

nicoburns: Looking at the Servo code, I'm pretty sure it's not. I've changed it to `recalculate_overflow`.

PR #253, opened 2025-10-10
Mark `border-color` and `border-radius` as repaint

nicoburns:
Servo PR: https://github.com/servo/servo/pull/39864


style/properties/longhands/border.mako.rs
```diff
@@ -27,6 +27,7 @@
         logical_group="border-color",
         allow_quirks="No" if is_logical else "Yes",
         ignored_when_colors_disabled=True,
+        servo_restyle_damage="repaint",
```
Loirooriol, line 30: For collapsed table borders, the color is resolved during layout. So I don't think this will work

nicoburns: Hmm... we should probably change that (presumably by resolving the *source* of the color in layout but not the actual color) because changing `border-color` of e.g. a button on `:hover` or `:active` is pretty common and we really don't want to be re-running layout for that.

PR #260, opened 2025-11-05
Allow updating DPI, viewport, and theme of a pre-existing `Device`

mrobinson:
This will allow Servo to update the `Device` and do media queries
without having to create a new one.

### DIFF
--- a/style/servo/media_queries.rs
+++ b/style/servo/media_queries.rs
@@ -225,6 +225,20 @@ impl Device {
         self.used_font_metrics.load(Ordering::Relaxed)
     }
 
+    /// Get the viewport size on this [`Device`].
+    pub fn viewport_size(&self) -> Size2D<f32, CSSPixel> {
+        self.viewport_size
+    }
+
+    /// Set the viewport size on this [`Device`].
+    ///
+    /// Note that this does not update any associated `Stylist`. For this you must call
+    /// `Stylist::media_features_change_changed_style` and
+    /// `Stylist::force_stylesheet_origins_dirty`.
+    pub fn set_viewport_size(&mut self, viewport_size: Size2D<f32, CSSPixel>) {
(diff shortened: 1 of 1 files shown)


mrobinson:
>  * Is the intention of these new functions to bypass that (because it's not needed for these use cases). Or is it required that the caller apply those invalidation manually when calling these functions

The intention isn't to bypass it, but delay it until the next layout. In the meantime, the device can be used for media queries which do not require the stylesheets to be dirtied. 

>  * If invalidation is required, should we make these function apply it automatically? and/or document correct usage of the function?

I'm not sure these function can apply it automatically, because the invalidation actually happens in the `Stylist` and not the the `Device`. In general the relationship between changing the `Device` (or replacing it) hasn't changed in the Stylo API. We are just doing it in a different place in Servo, so the semantics of this API don't really change. This is the invalidation in question: https://github.com/servo/servo/pull/40432/files#diff-49c9dea0cc9af0161b4681a4c4b1ed81f38f8190bd6854fd2ba6b3d74502b047R998

nicoburns (approved):
Looks good.

I think it would be nice to document the need for:

```rust
let sheet_origins_affected_by_device_change = stylist
      .media_features_change_changed_style(&guards, self.device());
stylist
    .force_stylesheet_origins_dirty(sheet_origins_affected_by_device_change);
```

on the `device_mut` method.

mrobinson:
> ```rust
> let sheet_origins_affected_by_device_change = stylist
>       .media_features_change_changed_style(&guards, self.device());
> stylist
>     .force_stylesheet_origins_dirty(sheet_origins_affected_by_device_change);
> ```
> 
> on the `device_mut` method.

Since `device_mut` is shared code, I've just put this documentation on the new setters. Hopefully that's okay.

PR #261, opened 2025-11-06
Enable text-overflow for servo

RichardTjokroutomo:
Enable `text-overflow` for Servo. Companion PR: [#40448](https://github.com/servo/servo/pull/40448)


Loirooriol:
You will need to rebase this, since the `main` branch was force-pushed

PR #268, opened 2025-11-14
Parse font-variation-settings inside @font-face rule

simonwuelker:


style/font_face.rs
```diff
@@ -576,7 +576,7 @@ impl Parse for Source {
 
 macro_rules! is_descriptor_enabled {
     ("font-variation-settings") => {
-        static_prefs::pref!("layout.css.font-variations.enabled")
```
simonwuelker, line 579: The problem here is that `layout.css.font-variations.enabled` is always false (and it's a `static_pref`, so we can't set it to true at runtime).

I'm not sure if gecko uses this code.

Either way, I'm open for alternative solutions (:

nicoburns: We set static prefs here: https://github.com/servo/stylo/blob/main/stylo_static_prefs/src/lib.rs. This code is not shared with Gecko (they have their own version which may set different prefs to us).

simonwuelker: I see, thank you

mrobinson (approved):

simonwuelker:
The corresponding servo PR was approved, can someone merge this?

Issue #223, opened 2025-08-05
Allow increasing stack size for use with ASAN


jschwe:
Allow configuring the stack size for the threadpool, to support running with ASAN enabled, which increases the necessary stack requirements. 

https://github.com/servo/stylo/blob/0d8dd5455daf3bc445312032a0a430e11a817390/style/parallel.rs#L38

PR #269, opened 2025-11-15
servo: Allow configuring style stack size

jschwe:
When running Servo with ASAN we hit the stack limit. We can use compile-time parsing of environment variables, to allow the embedder to configure the style thread stack size.

Fixes https://github.com/servo/stylo/issues/223

### DIFF
--- a/style/parallel.rs
+++ b/style/parallel.rs
@@ -34,8 +34,21 @@ pub const STYLE_THREAD_STACK_SIZE_KB: usize = 256;
 
 /// The minimum stack size for a thread in the styling pool, in kilobytes.
 /// Servo requires a bigger stack in debug builds.
+/// We allow configuring the size, since running with ASAN requires an even larger
+/// stack size.
 #[cfg(feature = "servo")]
-pub const STYLE_THREAD_STACK_SIZE_KB: usize = 512;
+pub const STYLE_THREAD_STACK_SIZE_KB: usize = const {
+    let default_stack_size = 512;
+    if let Some(user_def_size) = option_env!("SERVO_STYLE_THREAD_STACK_SIZE_KB") {
+        if let Ok(user_def_size) = usize::from_str_radix(user_def_size, 10) {
+            user_def_size
+        } else {
+            panic!("SERVO_STYLE_THREAD_STACK_SIZE_KB must be a valid integer")
+        }
+    } else {
+        default_stack_size
+    }
+};
 
 /// The stack margin. If we get this deep in the stack, we will skip recursive
 /// optimizations to ensure that there is sufficient room for non-recursive work.


style/parallel.rs
```diff
@@ -34,8 +34,21 @@ pub const STYLE_THREAD_STACK_SIZE_KB: usize = 256;
...
+    if let Some(user_def_size) = option_env!("SERVO_STYLE_THREAD_STACK_SIZE_KB") {
+        if let Ok(user_def_size) = usize::from_str_radix(user_def_size, 10) {
+            user_def_size
+        } else {
+            panic!("SERVO_STYLE_THREAD_STACK_SIZE_KB must be a valid integer")
+        }
+    } else {
+        default_stack_size
+    }
+};
```
mrobinson, line 51: Would it be possible to make this a bit simpler and just increase the stack size when ASAN is enabled?

jdm: Given https://github.com/servo/servo/issues/40814 I would also like to do this for debug builds, if possible.

jschwe: I would prefer keeping it configurable too, since it allows embedders to adjust the stack size based on their needs. 
The code is quite simple, just very verbose since we can't use any traits in const contexts, like `and_then()` or `unwrap_or()`.

jschwe: @mrobinson Coming back to this, could we revive the discussion here on how to continue?

mrobinson: > Given [servo/servo#40814](https://github.com/servo/servo/issues/40814) I would also like to do this for debug builds, if possible.

We have already increased the size here just for Servo due to our debug builds. I guess we could increase it again. If possible, it would be nice to have the build just work, no matter the build configuration, without having to set a magic environment variable. Also, what are the implications of exposing this in production builds?

Yet...why do we need a larger stack size than Gecko?

jschwe: > Also, what are the implications of exposing this in production builds?

The environment variable is read at compile-time, so I don't see any issue there. 

> We have already increased the size here just for Servo due to our debug builds. I guess we could increase it again. If possible, it would be nice to have the build just work, no matter the build configuration, without having to set a magic environment variable.

As I understand, the whole point of this setting is to reduce the amount of memory the stylo threads use (since AFAIK the default is higher, I don't remember exactly but either 2 or 8 MB?). Increasing works, but it since release / production builds will require less memory due to optimizing, it would also be wasteful to just choose the highest option. 

> Yet...why do we need a larger stack size than Gecko?

I wonder if they perhaps always optimize stylo to a certain degree? (pure speculation)

jschwe: > I would also like to do this for debug builds, if possible.

The difficult thing here is that we can't detect that. We can only check `debug-assertions`, but we also enable those in release mode in servo so it's not optimal. 

But I guess I could live with a `cfg(any(debug_assertions, asan))`, although I'm not a huge fan, if thats preferred.

jschwe: > The difficult thing here is that we can't detect that. We can only check debug-assertions, but we also enable those in release mode in servo so it's not optimal.

Thinking about it again, we need to ensure that we run servo CI with the smaller stack size, to prevent sudden crashes in production. And since we do want to enable debug_assertions in servos CI, those two goals kind of exclude themselves. So we are back to the embedder build system somehow needing to pass the information to stylo, which could either be a `cfg`, which would be a global rustflag, a feature flag "debug", or an environment variable like the current one.

mrobinson: Apparently, Gecko's default debug build does include some Rust optimizations. Maybe we could do that same.

jschwe: > Apparently, Gecko's default debug build does include some Rust optimizations. Maybe we could do that same.

I'm not opposed to that. IMHO it might make sense to optimize all non-workspace dependencies in servo with `1` or perhaps even `2`. It would slow down clean debug builds, improve the performance of debug builds, and perhaps slightly regress the ability to debug issues in non-workspace dependencies with gdb. Seems like a trade-off we could make if we document it properly in the book. 

Anyway, we would probably still need to increase the stack size for asan, although a global cfg option for that would be fine, since we anyway need to set rustflags for asan.

style/parallel.rs
```diff
@@ -34,8 +34,21 @@ pub const STYLE_THREAD_STACK_SIZE_KB: usize = 256;
...
+    let default_stack_size = 512;
+    if let Some(user_def_size) = option_env!("SERVO_STYLE_THREAD_STACK_SIZE_KB") {
+        if let Ok(user_def_size) = usize::from_str_radix(user_def_size, 10) {
+            user_def_size
+        } else {
+            panic!("SERVO_STYLE_THREAD_STACK_SIZE_KB must be a valid integer")
+        }
+    } else {
+        default_stack_size
+    }
```
mrobinson, line 50: Can `map` be used here?

jschwe: I'm pretty sure I tried that, and its not available in const contexts sadly. 
Edit: [`docs.rs`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map) confirms `map` on option is still unstable in const contexts.

jschwe:
FYI, I can't merge to this repo, so someone else would need to click the merge button if everything is alright otherwise.

PR #270, opened 2025-11-18
enable text-overflow property for Servo

RichardTjokroutomo:
Enable `text-overflow` for Servo. Companion PR: [#40526](https://github.com/servo/servo/pull/40526)


Loirooriol (changes requested):
As said in the Servo review, since it's not handling strings or 2 values, this should alter the parser to reject these cases for Servo.

Alternatively, you can put this property behind a flag, and then in Servo make it an experimental web platform feature flag. Then it won't be enabled by default, but it will be used in tests.

style/properties/longhands/text.mako.rs
```diff
@@ -9,9 +9,7 @@ ${helpers.predefined_type(
     "TextOverflow",
     "computed::TextOverflow::get_initial_value()",
     engines="gecko servo",
-    servo_pref="layout.unimplemented",
     animation_type="discrete",
-    boxed=True,
```
Loirooriol: Why remove this?

style/values/specified/text.rs
```diff
@@ -182,13 +182,14 @@ impl Parse for InitialLetter {
     ToResolvedValue,
     ToShmem,
 )]
-#[repr(C, u8)]
```
Loirooriol, line 177: Why remove this?

style/values/specified/text.rs
```diff
@@ -227,13 +228,22 @@ impl Parse for TextOverflow {
         context: &ParserContext,
         input: &mut Parser<'i, 't>,
     ) -> Result<TextOverflow, ParseError<'i>> {
-        let first = TextOverflowSide::parse(context, input)?;
+        let first = <TextOverflowSide as Parse>::parse(context, input)?;
```
Loirooriol: Why add this?

style/values/specified/text.rs
```diff
@@ -227,13 +228,22 @@ impl Parse for TextOverflow {
...
         Ok(
-            if let Ok(second) = input.try_parse(|input| TextOverflowSide::parse(context, input)) {
-                Self {
-                    first,
-                    second,
-                    sides_are_logical: false,
+            if let Ok(second) = input.try_parse(|input| <TextOverflowSide as Parse>::parse(context, input)) {
```
Loirooriol: Why add this?

style/values/specified/text.rs
```diff
@@ -227,13 +228,22 @@ impl Parse for TextOverflow {
...
+                if cfg!(feature = "servo") {
```
Loirooriol: This will just ignore the 1st value. Two values should instead be invalid. So don't parse twice.

Loirooriol (changes requested):

style/properties/longhands/text.mako.rs
RichardTjokroutomo: The following error:

```
error[E0080]: attempt to compute `0_usize - 1_usize`, which would overflow
     --> D:\stylo-bz\stylo\target\debug\build\stylo-e19c9c11b274e2bc\out/properties.rs:92246:1
      |
92246 | const_assert!(std::mem::size_of::<longhands::text_overflow::SpecifiedValue>() > BOX_THRESHOLD);     
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `properties::generated::_::{constant#0}` failed here
```

Is caused because when we use `Servo`, `String(crate::values::AtomString)` doesn't exist, so the total size of the box is too small, thus failing the check.

`boxed=True` is removed because of the following comment in [properties.mako.rs](https://github.com/servo/stylo/blob/main/style/properties/properties.mako.rs):

```
// There are two reasons for this test to fail:
//
//   * Your changes made a specified value type for a given property go
//     over the threshold. In that case, you should try to shrink it again
//     or, if not possible, mark the property as boxed in the property
//     definition.
//
//   * Your changes made a specified value type smaller, so that it no
//     longer needs to be boxed. In this case you just need to remove
//     boxed=True from the property definition. Nice job!
```

(credits to @minghuaw for finding the solution)

style/values/specified/text.rs
RichardTjokroutomo:
```
error[E0566]: conflicting representation hints
   --> style\values\specified\text.rs:185:8
    |
185 | #[repr(C, u8)]
    |        ^  ^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #68585 <https://github.com/rust-lang/rust/issues/68585>
    = note: `#[deny(conflicting_repr_hints)]` on by default
```

style/values/specified/text.rs
RichardTjokroutomo: So for some reason, stylo now can't correctly decide which `Parse` to use (or rather, for some reason in the past, it can choose the right one), and so if you don't specify, it will result in a compilation error:

```
error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> style\values\specified\text.rs:231:21
    |
231 |         let first = TextOverflowSide::parse(context, input)?;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^ ------- unexpected argument #1 of type `&ParserContext<'_>`
    |
note: associated function defined here
   --> style\values\specified\text.rs:178:5
    |
178 |     Parse,
    |     ^^^^^
    = note: this error originates in the derive macro `Parse` (in Nightly builds, run with -Z macro-backtrace for more info)
help: remove the extra argument
    |
231 -         let first = TextOverflowSide::parse(context, input)?;
231 +         let first = TextOverflowSide::parse(input)?;
```

style/values/specified/text.rs
RichardTjokroutomo: Previously, I tried to simply feature-gate the if block:

```rs
#[cfg(feature = "gecko")]
                Self {
                    first,
                    second,
                    sides_are_logical: false,
                }
```

but again, after the rebase the compiler is complaining mismatched type.

style/properties/longhands/text.mako.rs
Loirooriol: OK, then set it to `engine == 'gecko'`

Loirooriol (commented):
Since changing the struct affects things more than what I expected, maybe just do something like

```rs
        #[cfg(feature = "servo")]
        if matches!(first, TextOverflowSide::String(_)) {
            return Err(input.new_custom_error(StyleParseErrorKind::UnspecifiedError));
        }
        #[cfg(feature = "gecko")]
        if let Ok(second) = input.try_parse(|input| TextOverflowSide::parse(context, input)) {
            return Ok(Self {
                first,
                second,
                sides_are_logical: false,
            });
        }
        Ok(Self {
            first: TextOverflowSide::Clip,
            second: first,
            sides_are_logical: true,
        })
```

Or add support for strings as Nico says in https://github.com/servo/servo/pull/40526#issuecomment-3581285687

style/values/specified/text.rs
Loirooriol: What I'm saying is that `CSS.supports("text-overflow: ellipsis ellipsis")` should be false.

style/properties/longhands/text.mako.rs
RichardTjokroutomo: I guess this won't be necessary anymore

style/values/specified/text.rs
RichardTjokroutomo: As with the other case, the compiler now can't resolve to the correct `Parse`.

```
error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> style\values\specified\text.rs:231:21
    |
231 |         let first = TextOverflowSide::parse(context, input)?;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^ ------- unexpected argument #1 of type `&ParserContext<'_>`
    |
note: associated function defined here
   --> style\values\specified\text.rs:178:5
    |
178 |     Parse,
    |     ^^^^^
    = note: this error originates in the derive macro `Parse` (in Nightly builds, run with -Z macro-backtrace for more info)
help: remove the extra argument
    |
231 -         let first = TextOverflowSide::parse(context, input)?;
231 +         let first = TextOverflowSide::parse(input)?;
```

style/values/specified/text.rs
Loirooriol: Undo it, no longer needed

PR #283, opened 2025-12-27
Add layout.css.contrast-color.enabled to static prefs

webbeef:
This will enable the `contrast-color()` function in `<color>` productions: https://drafts.csswg.org/css-color-5/#contrast-color


Loirooriol (approved):
Please open a Servo PR to handle test expectations.

Also the description should mention what this pref does: this will enable the `contrast-color()` function in `<color>` productions. https://drafts.csswg.org/css-color-5/#contrast-color

PR #296, opened 2026-01-27
Add `:-servo-list-box` pseudo-class

lukewarlow:
This is for selects in the listbox state e.g. `<select size="3">`

Servo PR: https://github.com/servo/servo/pull/42194


stylo_dom/lib.rs
```diff
@@ -143,6 +143,8 @@ bitflags! {
         const ACTIVE_VIEW_TRANSITION = 1u64 << 50;
         /// For :-moz-suppress-for-print-selection.
         const SUPPRESS_FOR_PRINT_SELECTION = 1u64 << 51;
+        /// Non-standard & undocumented.
```
Loirooriol: This doesn't tell much about what it does. You can say that it's for selects in the listbox state, with a link to https://html.spec.whatwg.org/multipage/rendering.html#list-box

PR #342, opened 2026-03-23
Fix CSS step timing functions `jump-both`, `jump-start`, `start` at timestamp 0; Early return when possible

yezhizhen:
- The keyframe selection used <= for normal direction, causing the first keyframe to be selected at progress 0; changed to < for normal direction (kept <= for reverse to avoid https://github.com/servo/stylo/pull/342#discussion_r2975214718) and added direction‑aware step‑function handling. 
- Guard against division‑by‑zero for zero‑length keyframe intervals.

### Make early return happen earlier:
1. total_progress ≥ 1.0 and total_progress < 0 now handled immediately. Also make it direction aware.
2. At timestamp 0, for normal direction, non‑jump step functions (jump‑end, jump‑none, end) now return the start‑keyframe value early, instead of proceeding to interpolation.
3. Zero-interval: we exit early to avoid division‑by‑zero in `percentage_between_keyframes`

[Try](https://github.com/servo/servo/runs/68226116022)
Servo PR: https://github.com/servo/servo/pull/43594


style/servo/animation.rs
```diff
@@ -634,7 +634,7 @@ impl Animation {
                     .computed_steps
                     .iter()
                     .rev()
-                    .position(|step| total_progress as f32 <= 1. - step.start_percentage)
+                    .position(|step| (total_progress as f32) <= 1. - step.start_percentage)
```
Loirooriol: Should this be `<` too?

yezhizhen: Actually we shouldn't. Otherwise for reversed direction at timestamp 0, we jump while we shouldn't for `jump-start`, `jump-both`, `start`.

But there's more severe problem I've been trying to fix based on horrible test results..

style/servo/animation.rs
```diff
@@ -657,23 +670,36 @@ impl Animation {
...
             return;
         }
-        if total_progress >= 1.0 {
-            add_declarations_to_map(&next_keyframe);
-            return;
+
+        // At progress 0, we need to handle step functions specially
+        // for "jump-both, jump-start, start" step functions.
+        if total_progress == 0.0 {
+            if let TimingFunction::Steps(_steps, pos) = &prev_keyframe.timing_function {
```
Loirooriol, line 702: I tend to think this seems the wrong place to handle timing functions. Shouldn't it be handled in `calculate_step_output()`?

yezhizhen: It is handled there. But previously never get the chance to reach there (L723) due to early return.

The [test result](https://github.com/servo/servo/actions/runs/23441877600) looks decent.

There's only one failure case that I hope to check tomorrow.

yezhizhen: No more failure.
https://github.com/servo/servo/runs/68226116022

Would you like to review first? I will update PR details tomorrow.

yezhizhen (commented):
We also made early return happen earlier in various cases, instead of proceed to interpolation.

See "Make early return happen earlier" in description.

style/servo/animation.rs
```diff
@@ -649,31 +679,40 @@ impl Animation {
...
-            "Animation::get_property_declaration_at_time: keyframe from {:?} to {:?}",
-            prev_keyframe_index, next_keyframe_index
-        );
-
         let prev_keyframe = &self.computed_steps[prev_keyframe_index];
         let next_keyframe = match next_keyframe_index {
             Some(index) => &self.computed_steps[index],
-            None => return,
+            None => {
+                debug_assert!(false, "next_keyframe_index should always be Some");
```
yezhizhen, line 687: This is because:
1. We always have steps at 0% and 100% https://github.com/servo/stylo/blob/a4f1d69e3d239b73687ee5d6eca59d8e4056da41/style/servo/animation.rs#L174-L175
2. This always exists for `total_progress` in [0,1), no matter reverse or normal.

style/servo/animation.rs
```diff
@@ -591,15 +591,16 @@ impl Animation {
             return;
         }
 
-        let total_progress = match self.state {
+        // Raw progress ratio of the animation: can be negative (before start) or >1.0 (after end).
```
simonwuelker: nit: it can also be >1.0 while the animation is running, if `animation-iteration-count` is greater than 1.

style/servo/animation.rs
```diff
@@ -649,31 +679,40 @@ impl Animation {
...
-        // If we only need to take into account one keyframe, then exit early
-        // in order to avoid doing more work.
-        let mut add_declarations_to_map = |keyframe: &ComputedKeyframe| {
-            for value in keyframe.values.iter() {
-                map.insert(value.id().to_owned(), value.clone());
-            }
-        };
-        if total_progress <= 0.0 {
+        // Detect zero interval. Prevent division by zero from percentage_between_keyframes.
+        if Some(prev_keyframe_index) == next_keyframe_index {
```
simonwuelker: The % between two keyframes should never be zero, because if it was then `IntermediateComputedKeyframe::generate_from_keyframes` should have coalesced those two frames.

If you're seeing two keyframes with the same % then I suspect the bug is in that function and not here.

style/servo/animation.rs
yezhizhen: Indeed!

style/servo/animation.rs
yezhizhen: It is not a bug of that function, but the result of index computation.

For example, reverse direction at `total_progress == 0.0` can trigger the equality.

style/servo/animation.rs
xiaochengh: Agreed to Oriol. Btw I also find `PropertyAnimation` below weird, as it's resolving the timing function for every property. I imagine a better process would be:

1. Use `timing_function.calculate_output()` to resolve the progress to a `resolved_progress`, which is number in [0, 1]
2. Use `prev_keyframe` if `resolved_progress` is 0; Use `next_keyframe` is `resolved_progress` is 1
3. Otherwise, calculate a linear interpolation between the two keyframes at `resolved_progress`

In this way we don't need handle special handling of any timing function here.

yezhizhen: > Shouldn't it be handled in calculate_step_output()?

`fn calculate_step_output()` is shared with Firefox. But Firefox doesn't have the problem we have here :/
This PR only changes Servo specific code.

xiaochengh:
Looks like there will be some extended discussion regarding the step functions.

Can we split the "early return when possible" part into a different PR and land it first?

yezhizhen:
I will abandon this work as I don't know how to solve conflict.

Loirooriol:
Try `git rebase --onto main a4f1d69e3d239b73687ee5d6eca59d8e4056da41 fix-quantization`

yezhizhen:
> Try `git rebase --onto main a4f1d69e3d239b73687ee5d6eca59d8e4056da41 fix-quantization`

I don't know how to solve conflict with latest keyframe change: dcbc0da7ed1d73dab40db7c6ce3b27fbd21db2d9
Will find a time to understand that and resolve it.

simonwuelker:
> > Try `git rebase --onto main a4f1d69e3d239b73687ee5d6eca59d8e4056da41 fix-quantization`
> 
> I don't know how to solve conflict with latest keyframe change: [dcbc0da](https://github.com/servo/stylo/commit/dcbc0da7ed1d73dab40db7c6ce3b27fbd21db2d9) Will find a time to understand that and resolve it.

I'm happy to help out if you have any questions.

yezhizhen:
Resurrected in #354! Let me try some testings first tho :D

I'm unable to directly reopen this PR.

PR #345, opened 2026-04-01
Remove unused `PseudoElement::DetailsSummary`

### LINKED ISSUE #43812: Stylo: Remove `PseudoElement::DetailsSummary`
Servo is some asserts and such around that, but there is nothing implementing this pseudo-element.

In Stylo, remove https://github.com/servo/stylo/blob/8557228b96c0e343764953e72a62ea503baf01b3/style/servo/selector_parser.rs#L60

Here in Servo, remove `::-servo-details-summary` from stylesheets.

TG199:
Fixes servo/servo#43812

### DIFF
--- a/style/servo/selector_parser.rs
+++ b/style/servo/selector_parser.rs
@@ -57,7 +57,6 @@ pub enum PseudoElement {
 
     // Non-eager pseudos.
     Backdrop,
-    DetailsSummary,
     DetailsContent,
     Marker,
 
--- a/style/servo/selector_parser.rs
+++ b/style/servo/selector_parser.rs
@@ -98,7 +98,6 @@ impl ToCss for PseudoElement {
             Selection => "::selection",
             FirstLetter => "::first-letter",
             Backdrop => "::backdrop",
-            DetailsSummary => "::-servo-details-summary",
             DetailsContent => "::details-content",
             Marker => "::marker",
             ColorSwatch => "::color-swatch",
@@ -256,7 +255,6 @@ impl PseudoElement {
             PseudoElement::Backdrop
             | PseudoElement::ColorSwatch
             | PseudoElement::FileSelectorButton
-            | PseudoElement::DetailsSummary
             | PseudoElement::Marker
             | PseudoElement::Placeholder
             | PseudoElement::DetailsContent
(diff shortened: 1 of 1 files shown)


Loirooriol (changes requested):
Please ensure the patch compiles (`cargo build`)

Loirooriol:
Also, once this compiles, please prepare a companion PR for Servo where you change `Cargo.toml` so that the Stylo crates point to this change (use `rev = "refs/pull/345/head"`). Ensure that compiles, and also remove the useless `::-servo-details-summary` from UA stylesheets.

nicoburns:
I know that this is technically correct, but I'm seeing a lot of breakage in the real world from unimplmented psuedo elements. The problem is that the parser will drop the entire selector if it doesn't recognise any part of it, so if someone uses the `,` to target a basic selector as well as a pseudo (common in CSS resets and other such "base styles") then the style will fail to apply even to the other elements.

What is the actual hazard from being a bit loose around parsing extra pseudos? Presumably they won't ever match anything anyway? So having them parse seems harmless? (assuming they're standard pseudos which will parse correctly in other browsers)

jdm:
> I know that this is technically correct, but I'm seeing a lot of breakage in the real world from unimplmented psuedo elements. The problem is that the parser will drop the entire selector if it doesn't recognise any part of it, so if someone uses the `,` to target a basic selector as well as a pseudo (common in CSS resets and other such "base styles") then the style will fail to apply even to the other elements.
> 
> What is the actual hazard from being a bit loose around parsing extra pseudos? Presumably they won't ever match anything anyway? So having them parse seems harmless? (assuming they're standard pseudos which will parse correctly in other browsers)

That seems like something worth discussing on zulip instead of this PR to remove an unused, servo-internal pseudo.

nicoburns:
> That seems like something worth discussing on zulip instead of this PR to remove an unused, servo-internal pseudo.

Agree on Zulip.

> servo-internal pseudo.

Oh, huh. I thought this was standardised (because details-content is). But now that I think about it, it's a bit odd to have a details-summary pseudo element, because `<summary>` is an actual element!

TG199:
> Please ensure the patch compiles (`cargo build`)

Hi @Loirooriol, following the recent comments on this PR, should I continue with this or wait till the discussion is resolved. 
The patch compiles now but running `cargo clippy` produces over 2k warnings, should I be bothered about that?

Loirooriol:
@TG199 `::-servo-details-summary` isn't standard and I doubt any website is using it, so this PR is not affected by the discussion of the parser accepting pseudo-elements that we don't actually support.

> The patch compiles now

Please push the new version.

> cargo clippy produces over 2k warnings

I guess we are not running clippy in Stylo. Presumably it should be the same without your patch.

Loirooriol (commented):
Please undo all the unrelated formatting changes.

Loirooriol (commented):
Looking good, can you create the Servo companion PR?

TG199:
> Looking good, can you create the Servo companion PR?

Sure, on it.

Loirooriol (approved):

PR #350, opened 2026-04-09
 Stylo: Enable text-align: match-parent for Servo

TG199:
Removes the `#[cfg(feature = "gecko")]` gate on `TextAlign::MatchParent`
in `style/values/specified/text.rs`, enabling `text-align: match-parent`
for Servo. The computation logic (resolving `start`/`end` to physical
directions based on parent writing mode) was already correct and needed
no changes.

Servo PR: https://github.com/servo/servo/pull/44073


Loirooriol (approved):
Please post a companion PR for Servo

PR #351, opened 2026-04-10
Enable @container rule for servo

BudiArb:
Enable @container rule for servo

Servo PR: servo/servo#44083


style/stylesheets/rule_parser.rs
```diff
@@ -719,7 +719,7 @@ impl<'a, 'i> AtRuleParser<'i> for NestedRuleParser<'a, 'i> {
             "font-face" => {
                 AtRulePrelude::FontFace
             },
-            "container" if cfg!(feature = "gecko") => {
+            "container" => {
```
Loirooriol, line 722: While you work on it, it may be good to keep it behind an experimental pref, like

```rs
"container" if container_enabled() => {
```

```rs
#[cfg(feature = "gecko")]
fn container_enabled() {
    true
}

#[cfg(feature = "servo")]
fn container_enabled() {
    static_prefs::pref!("layout.container-queries.enabled")
}
```

PR #354, opened 2026-04-20
Correctly compute keyframe at timestamp 0; Early return when possible

yezhizhen:
- The keyframe selection used <= for normal direction, causing the first keyframe to be selected at progress 0; changed to < for normal direction (kept <= for reverse to avoid https://github.com/servo/stylo/pull/342#discussion_r2975214718) and added direction‑aware step‑function handling. 
- Guard against division‑by‑zero for zero‑length keyframe intervals.

### Make early return happen earlier:
1. total_progress == 1.0 and total_progress < 0 now handled immediately. Also make it direction aware.
2. Zero-interval: we exit early to avoid division‑by‑zero in `percentage_between_keyframes`

[Try](https://github.com/servo/servo/runs/68226116022)
Servo PR: https://github.com/servo/servo/pull/44365

### DIFF
--- a/style/servo/animation.rs
+++ b/style/servo/animation.rs
@@ -27,7 +27,7 @@ use crate::stylesheets::keyframes_rule::{KeyframesAnimation, KeyframesStep, Keyf
 use crate::stylesheets::layer_rule::LayerOrder;
 use crate::values::animated::{Animate, Procedure};
 use crate::values::computed::TimingFunction;
-use crate::values::generics::easing::BeforeFlag;
+use crate::values::generics::easing::{BeforeFlag, StepPosition};
 use crate::values::specified::TransitionBehavior;
 use crate::Atom;
 use parking_lot::RwLock;
@@ -756,7 +756,9 @@ impl Animation {
             return;
         }
 
-        let total_progress = match self.state {
+        // Raw progress ratio of the animation: can be negative (before start) or
+        // >1.0 (after end or during multiple iterations).
+        let progress = match self.state {
             AnimationState::Running | AnimationState::Pending | AnimationState::Finished => {
                 (now - self.started_at) / self.duration
             },
@@ -764,7 +766,7 @@ impl Animation {
             AnimationState::Canceled => return,
         };
 
-        if total_progress < 0.
+        if progress < 0.
             && self.fill_mode != AnimationFillMode::Backwards
             && self.fill_mode != AnimationFillMode::Both
         {
@@ -776,9 +778,41 @@ impl Animation {
         {
             return;
         }
-        let total_progress = total_progress
-            .min(self.current_iteration_end_progress())
-            .max(0.0);
+
+        // If we only need to take into account one keyframe, then exit early
+        // in order to avoid doing more work.
+        let mut add_declarations_to_map = |keyframe: &ComputedKeyframe| {
+            for value_or_reference in keyframe.values.iter() {
+                let AnimationValueOrReference::AnimationValue(value) = value_or_reference else {
+                    unreachable!("First or last keyframes define all properties");
+                };
+                map.insert(value.id().to_owned(), value.clone());
+            }
+        };
+
+        // Handle negative progress (before animation start) with backwards/both fill mode
+        if progress < 0.0 {
+            let keyframe = match self.current_direction {
+                AnimationDirection::Normal => self.computed_steps.first().unwrap(),
+                AnimationDirection::Reverse => self.computed_steps.last().unwrap(),
+                _ => unreachable!(),
+            };
+            add_declarations_to_map(keyframe);
+            return;
+        }
+
+        // Progress clamped to the current iteration (0.0 to 1.0).
+        let total_progress = progress.min(self.current_iteration_end_progress()).max(0.0);
+
+        if total_progress >= 1.0 {
+            let keyframe = match self.current_direction {
+                AnimationDirection::Normal => self.computed_steps.last().unwrap(),
+                AnimationDirection::Reverse => self.computed_steps.first().unwrap(),
+                _ => unreachable!(),
+            };
+            add_declarations_to_map(keyframe);
+            return;
+        }
 
         // Get the indices of the previous (from) keyframe and the next (to) keyframe.
         let next_keyframe_index;
@@ -789,7 +823,7 @@ impl Animation {
                 next_keyframe_index = self
                     .computed_steps
                     .iter()
-                    .position(|step| total_progress as f32 <= step.start_percentage);
+                    .position(|step| (total_progress as f32) < step.start_percentage);
                 prev_keyframe_index = next_keyframe_index
                     .and_then(|pos| if pos != 0 { Some(pos - 1) } else { None })
                     .unwrap_or(0);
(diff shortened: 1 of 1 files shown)


yezhizhen:
@Loirooriol Just friendly reminder.

style/servo/animation.rs
```diff
@@ -819,40 +853,42 @@ impl Animation {
             prev_keyframe_index, next_keyframe_index
         );
 
+        let prev_keyframe = &self.computed_steps[prev_keyframe_index];
         let Some(next_keyframe_index) = next_keyframe_index else {
+            debug_assert!(false, "next_keyframe_index should always be Some");
```
Loirooriol: Using `debug_unreachable` may be clearer.
Also the message isn't providing much, it should rather explain why.

style/servo/animation.rs
```diff
@@ -776,9 +778,41 @@ impl Animation {
...
+                _ => unreachable!(),
+            };
+            add_declarations_to_map(keyframe);
+            return;
+        }
+
+        // Progress clamped to the current iteration (0.0 to 1.0).
+        let total_progress = progress.min(self.current_iteration_end_progress()).max(0.0);
+
+        if total_progress >= 1.0 {
```
Loirooriol: Why can this handle the equality, but above you check `progress < 0.0`?

style/servo/animation.rs
yezhizhen: Done in https://github.com/servo/stylo/pull/354/commits/270133208fa27f2722d774b6a9f83f0b51b19c02

The explanation is a bit lengthy tho. We are relying on https://github.com/servo/stylo/blob/40aa59d7f3b2bce517c480c95a80d639843e1449/style/servo/animation.rs#L175-L176

style/servo/animation.rs
yezhizhen: In https://github.com/servo/stylo/pull/354/commits/270133208fa27f2722d774b6a9f83f0b51b19c02 I updated to `==1.0` as it is always in `[0,1]`.

> but above you check `progress < 0.0`?

At `progress == 0.0`, we don't want to early return because we need to reach the step‑function handling block at L878 we added.

Loirooriol: But don't we have the same jumping problem at the end?

yezhizhen: https://github.com/servo/servo/issues/43086

End has always been fine. We only have problem at the beginning.

yezhizhen:
@Loirooriol Do you still have other concerns for this?

xiaochengh (commented):
My previous comments at #342 seem unaddressed:
- https://github.com/servo/stylo/pull/342#discussion_r3008387316
- https://github.com/servo/stylo/pull/342#issuecomment-4168231514

yezhizhen:
I have removed special casing for step functions in https://github.com/servo/stylo/pull/354/commits/4c6b772360b1c0dd019aa122122a5f9eff91ba1a. It is no longer imported.

This should be much more straightforward to review now. We do miss one early return, but the readability is worth it.

Now the PR is mostly renaming, reordering, adding comments, and computing index at boundary correctly.

yezhizhen:
@Loirooriol Xiaocheng would probably be off this week. Do you mind checking this again?

style/servo/animation.rs
```diff
@@ -819,40 +855,27 @@ impl Animation {
...
-            }
-            return;
-        }
-        if total_progress >= 1.0 {
-            if reversed {
-                add_declarations_to_map(0);
-            } else {
-                add_declarations_to_map(self.computed_steps.len() - 1);
-            }
+        // Prevent division by zero from percentage_between_keyframes.
```
xiaochengh, line 873: Do we need to verify this with a test case?

xiaochengh (approved):
The implementation change looks very reasonable to me.

style/servo/animation.rs
yezhizhen: https://github.com/servo/servo/pull/44365#issuecomment-4279007889

Simon also asked same question. One test got something like 17895697.05px before, that's why I added this.

yezhizhen:
@mrobinson @Loirooriol I don't have write permission. Can you help merge this?

yezhizhen:
@Loirooriol @mrobinson I don't mean to nag. This isn't rocket science, but has been pended for two months. I also asked @Loirooriol two weeks ago if you still have concerns but just getting ignored.

xiaochengh:
Nit: maybe we need to modify the PR title, since the special handling for step functions have been removed?

style/servo/animation.rs
```diff
@@ -776,9 +779,42 @@ impl Animation {
...
+                map.insert(value.id().to_owned(), value.clone());
+            }
+        };
+
+        // Handle negative progress (before animation start) with backwards/both fill mode
+        if progress < 0.0 {
+            let keyframe = match self.current_direction {
+                AnimationDirection::Normal => self.computed_steps.first().unwrap(),
+                AnimationDirection::Reverse => self.computed_steps.last().unwrap(),
+                _ => unreachable!(),
```
mrobinson:
```suggestion
                _ => unreachable!("Current animation direction should never be Alternate."),
```

style/servo/animation.rs
```diff
@@ -776,9 +779,42 @@ impl Animation {
...
+
+        // Progress clamped to the current iteration [0.0, 1.0].
+        let total_progress = progress.min(self.current_iteration_end_progress()).max(0.0);
+
+        // At 1.0 there is nothing left to interpolate. Return end keyframe.
+        if total_progress == 1.0 {
+            let keyframe = match self.current_direction {
+                AnimationDirection::Normal => self.computed_steps.last().unwrap(),
+                AnimationDirection::Reverse => self.computed_steps.first().unwrap(),
+                _ => unreachable!(),
```
mrobinson:
```suggestion
                _ => unreachable!("Current animation direction should never be Alternate."),
```

style/servo/animation.rs
```diff
@@ -776,9 +779,42 @@ impl Animation {
...
+        };
+
+        // Handle negative progress (before animation start) with backwards/both fill mode
+        if progress < 0.0 {
+            let keyframe = match self.current_direction {
+                AnimationDirection::Normal => self.computed_steps.first().unwrap(),
+                AnimationDirection::Reverse => self.computed_steps.last().unwrap(),
+                _ => unreachable!(),
+            };
+            add_declarations_to_map(keyframe);
```
mrobinson: Let's avoid `unwrap()` here and also below:

```suggestion
            if let Some(keyframe) = match self.current_direction {
                AnimationDirection::Normal => self.computed_steps.first(),
                AnimationDirection::Reverse => self.computed_steps.last(),
                _ => unreachable!(),
            } {
                add_declarations_to_map(keyframe);
            }
```

style/servo/animation.rs
```diff
@@ -819,40 +855,27 @@ impl Animation {
...
+        // This can happen for reverse direction at total_progress == 0.0.
+        if prev_keyframe_index == next_keyframe_index {
+            add_declarations_to_map(&prev_keyframe);
             return;
         }
 
+        let reversed = self.current_direction != AnimationDirection::Normal;
+
         // Interpolate a new value for each animating property
         for property_index in 0..self.number_of_animating_properties {
```
mrobinson:
```suggestion
// Interpolate a new value for each animating property
let reversed = self.current_direction != AnimationDirection::Normal;
for property_index in 0..self.number_of_animating_properties {
```

mrobinson (approved):
Looks good with the nits below:

yezhizhen:
@mrobinson Thanks for review! Addressing this soon.

style/servo/animation.rs
yezhizhen: https://github.com/servo/stylo/blob/a683655d35822dc231598ac67c9343df33c48c5d/style/servo/animation.rs#L173-L176

But the animation is guaranteed to have them. We also unwrapped earlier in previous code.

yezhizhen (commented):
I have replaced the unreachable string as "Current animation direction can only be `normal` or `reverse`" instead, to be consistent with https://github.com/servo/stylo/blob/a683655d35822dc231598ac67c9343df33c48c5d/style/servo/animation.rs#L557-L558. Also added it to `fn iterate`.

style/servo/animation.rs
mrobinson: Still we should avoid unwrap when possible (https://github.com/servo/servo/issues/40744). If the code above changes, the person changing it might not get to this code, which can cause a panic later on.

yezhizhen: Good point! I guess there's nothing we can do with `unwrap` in this `debug_assert`.
Done in https://github.com/servo/stylo/pull/354/commits/01ef6d678bb10cf1a8f64bb985f6f269cfd2e3b1

PR #356, opened 2026-04-29
Enable the `user-select` property (as layout.unimplemented)

nicoburns:
I would like to use this property in Blitz. So if others don't mind, it would be helpful it could enabled for Servo builds of Stylo (gated behind `layout.unimplemented` as Servo doesn't implement it yet).


Loirooriol (approved):

Loirooriol:
There shouldn't be any problem, but would be good to bump Stylo in Servo just in case.

PR #362, opened 2026-05-12
Unify Servo's `SharedRwLock` implementation with Gecko

mrobinson:
Servo no longer uses the unsynchronized animation setup that's described
in the comment and `AtomicRefCell` is faster than `RwLock`. This change
has Servo use the same locking design that Gecko uses in order to
improve the performance of style locking / unlocking.

Servo PR: servo/servo#44883

### DIFF
--- a/style/shared_lock.rs
+++ b/style/shared_lock.rs
@@ -5,46 +5,33 @@
 //! Different objects protected by the same lock
 
 use crate::stylesheets::Origin;
-#[cfg(feature = "gecko")]
 use atomic_refcell::{AtomicRef, AtomicRefCell, AtomicRefMut};
-#[cfg(feature = "servo")]
-use parking_lot::RwLock;
 use servo_arc::Arc;
 use std::cell::UnsafeCell;
 use std::fmt;
-#[cfg(feature = "servo")]
-use std::mem;
-#[cfg(feature = "gecko")]
 use std::ptr;
 use style_traits::{CssString, CssStringWriter};
 use to_shmem::{SharedMemoryBuilder, ToShmem};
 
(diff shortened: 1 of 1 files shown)


style/shared_lock.rs
```diff
@@ -5,46 +5,33 @@
...
-    #[cfg_attr(feature = "servo", ignore_malloc_size_of = "Arc")]
-    arc: Arc<RwLock<()>>,
-
-    #[cfg(feature = "gecko")]
     cell: Option<Arc<AtomicRefCell<SomethingZeroSizedButTyped>>>,
 }
 
-#[cfg(feature = "gecko")]
+#[cfg(feature = "servo")]
+malloc_size_of::malloc_size_of_is_0!(SharedRwLock);
```
SimonSapin, line 32: I don’t think this is accurate? `SharedRwLock` can contain an `Arc`

mrobinson: The issue here is that `AtomicRefCell` doesn't implement `MallocSizeOf` so the resulting value is zero. In addition the thing that it holds inside is zero-sized as well. I'm not sure how to produce a reasonable value here.

nicoburns: There is always the option of just not implementing it, and then skipping the fields at the level above.

nicoburns (approved):
I tried this a few months back. I didn't notice any performance improvements, but I also didn't notice any issues. So it seems worth it just to unify the implementations.

style/shared_lock.rs
mrobinson: Hrm. In this case it would require annotating lots of call sites in Servo, so hopefully it is okay to keep this marked as 0 size?

PR #363, opened 2026-05-13
enable :has() selector

yezhizhen:
I have a scenario that is bugged by this.
Servo PR: https://github.com/servo/servo/pull/44902


Loirooriol:
When @mrobinson tried this, invalidation was not working

nicoburns:
Yeah, `:has` requires custom invalidation logic on the Servo side.

yezhizhen:
> Yeah, `:has` requires custom invalidation logic on the Servo side.

Thanks. Do we have a rough idea how hard is it to support this, and where we can start from in Servo side?

nicoburns:
> > Yeah, `:has` requires custom invalidation logic on the Servo side.
> 
> Thanks. Do we have a rough idea how hard is it to support this, and where we can start from in Servo side?

It's probably one of the hardest CSS feature to support! But it's definitely doable. The main trouble is that none of the requirements are documented, so it'd be a case of cribbing from the Gecko codebase and asking for help from Gecko developers (#layout in Mozilla matrix).

PR #364, opened 2026-05-13
Enable font-variant-ligatures for Servo

simonwuelker:
Companion PR for https://github.com/servo/servo/pull/44903


style/properties/shorthands.toml
```diff
@@ -439,14 +439,14 @@ sub_properties = [
   "font-family",
   "font-optical-sizing",
   "font-variation-settings",
-  "font-kerning"
+  "font-kerning",
+  "font-variant-ligatures"
```
mrobinson: Do you mind sorting this list before landing?

mrobinson (approved):

style/properties/shorthands.toml
simonwuelker: Sure

style/properties/shorthands.rs
```diff
@@ -2519,11 +2519,11 @@ pub mod font {
     use crate::properties::longhands::{
         font_family, font_feature_settings, font_language_override, font_size,
         font_size_adjust, font_variant_alternates, font_variant_east_asian, font_variant_emoji,
-        font_variant_ligatures, font_variant_numeric, font_variant_position,
+        font_variant_numeric, font_variant_position,
     };
     use crate::properties::longhands::{
-        font_kerning, font_optical_sizing, font_stretch, font_style, font_variant_caps, font_variation_settings,
-        font_weight,
+        font_kerning, font_variant_ligatures, font_optical_sizing, font_stretch, font_style, font_variant_caps,
```
Loirooriol: Please sort alphabetically

simonwuelker:
https://github.com/servo/servo/pull/44903 is approved, so this can be merged by someone with the relevant permissions.

style/properties/shorthands.toml
```diff
@@ -430,23 +430,23 @@ spec = "https://drafts.csswg.org/css-text-decor/#propdef-text-decoration"
...
   "font-optical-sizing",
+  "font-size",
+  "font-stretch",
+  "font-style",
+  "font-variant-caps",
+  "font-variant-ligatures",
   "font-variation-settings",
-  "font-kerning"
+  "font-weight",
+  "line-height"
```
Loirooriol: I don't think it's much important, but note the previous order seemingly came from the canonical order of the values (https://drafts.csswg.org/css-fonts-4/#font-prop), and it's webexposed.

style/properties/shorthands.rs
```diff
@@ -2878,7 +2874,6 @@ pub mod font_variant {
             .try_parse(|input| input.expect_ident_matching("none"))
             .is_ok()
         {
-            #[cfg(feature = "gecko")]
             {
                 ligatures = Some(FontVariantLigatures::NONE);
             }
```
Loirooriol: You can now remove the block wrapper, its purpose was avoiding `error[E0658]: attributes on expressions are experimental`

Loirooriol (approved):

style/properties/shorthands.toml
mrobinson: Oh, apologies for the suggestion. I guess we should keep following this order.

Loirooriol: Moving from `extra_gecko_sub_properties` to `sub_properties` will also change the order in Gecko anyways, so I think it's fine either way, but yeah maybe better to keep the change smaller

mrobinson (approved):

PR #365, opened 2026-05-14
Allow some `AttrValue` variants to be serialized lazily

mrobinson:
This allows some `AttrValue` variants, the ones that are easily
serializable and are sometimes set without existing serializations, to
be lazily serialized. The goal here is to allow storing these values,
but only serializing when necessary.

Servo PR: servo/servo#44931

### DIFF
--- a/malloc_size_of/lib.rs
+++ b/malloc_size_of/lib.rs
@@ -277,6 +277,14 @@ impl<T: MallocSizeOf> MallocSizeOf for std::cell::RefCell<T> {
     }
 }
 
+impl<T: MallocSizeOf> MallocSizeOf for std::sync::OnceLock<T> {
+    fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
+        self.get()
+            .map(|value| value.size_of(ops))
+            .unwrap_or_default()
+    }
+}
+
 impl<'a, B: ?Sized + ToOwned> MallocSizeOf for std::borrow::Cow<'a, B>
 where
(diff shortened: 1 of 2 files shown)


nicoburns:
What's the rationale for not also implementing this for `LengthPercentage`, `Color`, and `Dimension`?

mrobinson:
> What's the rationale for not also implementing this for `LengthPercentage`, `Color`, and `Dimension`?

Two reasons:

1. Currently they are all created via parsing a serialization, so we already have a serialization when they are created (which should be preserved).
2. There are no pre-existing serialization methods for these types, so they would have to be written -- but would be dead code due to number 1.

malloc_size_of/lib.rs
```diff
@@ -277,6 +277,12 @@ impl<T: MallocSizeOf> MallocSizeOf for std::cell::RefCell<T> {
     }
 }
 
+impl<T: MallocSizeOf> MallocSizeOf for std::sync::OnceLock<T> {
+    fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
+        self.get().size_of(ops)
```
SimonSapin: I think this might call `<Option<&T>>::size_of` which is always zero because of the `&`, but it should optionally call `T::size_of` which may not be zero

mrobinson: I've addressed this in the latest push.

SimonSapin (approved):

PR #367, opened 2026-05-15
Convert Servo's preferences into statically access atomics

mrobinson:
Static preferences are accessed by Stylo in very hot parts of restyling,
so this change exposes access to these preferences via `AtmicBool` and
`AtomicI32` instead of via `RwLock<HashMap<...>>`. This eliminates a
good deal of hashing work from profiles of layout hammering benchmarks.
In general, removing this work from styling will have beneficial
effects.

Now preferences are defined via a TOML file which is used to generate a
store of atomic primitives which are controlled via two macros.

Co-authored-by: Oriol Brufau <obrufau@igalia.com>


Loirooriol (approved):

style/properties/properties.mako.rs
```diff
@@ -383,30 +383,19 @@ impl NonCustomPropertyId {
...
-                let pref = match PREF_NAME[self.0 as usize] {
-                    None => return true,
-                    Some(pref) => pref,
-                };
-
-                // The assertions above guarantee that the pref defaults to false.
-                static_prefs::Preference::get(pref, false)
+                            vector.push(true);
+                        % endif %
+                % endfor
```
Loirooriol: The indentation is wrong

style/properties/properties.mako.rs
```diff
@@ -383,30 +383,19 @@ impl NonCustomPropertyId {
             % if engine == "gecko":
                 unsafe { structs::nsCSSProps_gPropertyEnabled[self.0 as usize] }
             % else:
-                static PREF_NAME: [Option<&str>; ${
-                    len(data.longhands) + len(data.shorthands) + len(data.all_aliases())
-                }] = [
+                static PREFERENCES: std::sync::LazyLock<Vec<bool>> = std::sync::LazyLock::new(|| {
```
Loirooriol: Would it be better to use an array?

```rs
                <% all_properties = data.longhands + data.shorthands + data.all_aliases() %>
                static PREFERENCES: std::sync::LazyLock<[bool; ${len(all_properties)}]> = std::sync::LazyLock::new(|| [
                % for property in all_properties:
                    <% preference = getattr(property, "servo_pref") %>
                    % if preference:
                        static_prefs::pref!("${preference}"),
                    % else:
                        true,
                    % endif %
                % endfor
                ]);
```

style/properties/properties.mako.rs
mrobinson: I've fixed this.

style/properties/properties.mako.rs
mrobinson: I was a bit worried about a static array size limit, but it seems to compile so we can try this.

PR #371, opened 2026-05-19
Respond to changes in `animation-delay`

yezhizhen:
`update_from_other` unconditionally restored the old `started_at`, ignoring the new delay.

To fix this, computed the correct `started_at` with the new delay.
We also compute the fresh new state based on it.

Servo companion: https://github.com/servo/servo/pull/45013

### DIFF
--- a/style/servo/animation.rs
+++ b/style/servo/animation.rs
@@ -707,7 +707,8 @@ impl Animation {
         self.current_direction = old_direction;
 
         if self.delay != old_delay {
-            // Shift by the delay delta:
+            // `started_at` incorporates the delay, so changing the delay necessarily changes `started_at`.
+            // Note: `started_at` may actually be in the future.
             self.started_at = old_started_at + (self.delay - old_delay);
 
             match old_state {
--- a/style/servo/animation.rs
+++ b/style/servo/animation.rs
@@ -697,57 +697,101 @@ impl Animation {
         // NB: We shall not touch the started_at field, since we don't want to
         // restart the animation.
         let old_started_at = self.started_at;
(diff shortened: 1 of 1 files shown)


mrobinson:
@yezhizhen Do you mind writing a bit more about what this change does in the PR description? I think it is useful for people reading the repository history.

yezhizhen:
I made a mistake. This only works for static ones. We cannot use the new `now`, as that would teleport the running animation back to beginning with the new delay.

The new passing WPT are all static ones.

yezhizhen:
Fixed. Now it works correctly on running ones.
Also get one more passing case:
`/css/css-transitions/all-interpolates-same-as-explicit-property.html`

Still recommend "Hide Whitespace diff"

style/servo/animation.rs
```diff
@@ -697,57 +697,101 @@ impl Animation {
...
-        match (&mut self.iteration_state, old_iteration_state) {
-            (
-                &mut KeyframesIterationState::Finite(ref mut iters, _),
-                KeyframesIterationState::Finite(old_iters, _),
-            ) => *iters = old_iters,
-            _ => {},
-        }
+        if self.delay != old_delay {
+            // Shift by the delay delta:
+            self.started_at = old_started_at + (self.delay - old_delay);
```
mrobinson:
```suggestion
            // `started_at` incorporates the delay, so changing the delay necessarily changes `started_at`.
            // Note: `started_at` may actually be in the future.
            self.started_at = old_started_at + (self.delay - old_delay);
```

mrobinson (approved):

PR #377, opened 2026-06-02
Support `font-variant-alternates` and `@font-feature-values` for servo

simonwuelker:
Servo PR: https://github.com/servo/servo/pull/45308


jdm (approved):

style/values/specified/font.rs
```diff
@@ -1138,6 +1138,7 @@ bitflags! {
     ToShmem,
 )]
 #[repr(C, u8)]
+#[cfg_attr(feature = "servo", derive(Deserialize, Hash, Serialize))]
```
Loirooriol: Don't restrict this to Servo

style/values/specified/font.rs
```diff
@@ -1177,12 +1178,23 @@ pub enum VariantAlternates {
 )]
 #[repr(transparent)]
 #[typed(todo_derive_fields)]
+#[cfg_attr(feature = "servo", derive(Deserialize, Hash, Serialize))]
```
Loirooriol: Ditto

style/values/mod.rs
```diff
@@ -546,6 +546,7 @@ impl<A: Debug, B: Debug> Debug for Either<A, B> {
     ToShmem,
 )]
 #[repr(C)]
+#[cfg_attr(feature = "servo", derive(Deserialize, Serialize))]
```
Loirooriol: Ditto

style/values/specified/font.rs
simonwuelker: We should probably un-gate these derives in other places as well then, I copied this from style types like

https://github.com/servo/stylo/blob/a64923b5d5c67313c81c5056f5e30ec0babb04d6/style/values/specified/box.rs#L865-L881

Loirooriol: Yes, done upstream: https://phabricator.services.mozilla.com/D311004

simonwuelker:
The servo PR is approved, so this can be merged now.

PR #378, opened 2026-06-02
Bump stylo to 0.18.0

jschwe:
d7e4d965d515b55b8d25a7344aa06f9bee4f0f2f is the version servo's 0.3.0 release depends on, so we can just bump stylo on main.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ members = [
 default-members = ["style"]
 
 [workspace.package]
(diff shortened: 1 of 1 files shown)


Loirooriol (approved):

nicoburns (approved):

jschwe:
Fyi, I don't have permissions to merge and/or release, so someone else will need to do that.

PR #379, opened 2026-06-02
Fix typo in `selectors` README

VladNegara:
I noticed a typo in the README file of the `selectors` library.

### DIFF
--- a/selectors/README.md
+++ b/selectors/README.md
@@ -6,7 +6,7 @@ rust-selectors
 * [crates.io](https://crates.io/crates/selectors)
 
 CSS Selectors library for Rust.
-Includes parsing and serilization of selectors,
+Includes parsing and serialization of selectors,
 as well as matching against a generic tree of elements.
 Pseudo-elements and most pseudo-classes are generic as well.
 


Loirooriol (commented):
Since this isn't specific to Servo, can you send this directly to upstream Firefox?

If it lands here, I will need to export it anyways (which isn't a big deal if you aren't familiar with Firefox contributions)

VladNegara:
I'm not familiar with Firefox contributions. If the process isn't too involved, could you point me in the right direction?

Loirooriol:
See https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html for a guide.

In this case you don't need to build. Also downloading the entire history would be overkill, I think `git clone --depth 1 https://github.com/mozilla-firefox/firefox/` should suffice to get the code.

With the code, just do the change, and commit. You should probably link the commit to a bug in https://bugzilla.mozilla.org/, e.g. using `Bug XXXXXX - Fix typo in selectors README. r?#style` as a commit title.

Then submit it according to https://firefox-source-docs.mozilla.org/contributing/contribution_quickref.html#to-submit-a-patch. 

If that's too tedious, it's fine to just land it here.

Loirooriol (approved):

PR #382, opened 2026-06-04
style: Decode UTF8 data in FontTag within Debug impl

simonwuelker:
`FontTag` contains a 4 byte tag that is *usually* valid UTF8. For debugging purposes its more convenient to display it as a string instead of a raw `u32`.

Before: `FontTag(1818847073)`
After: `FontTag("liga")`


nicoburns (approved):

mrobinson (approved):

simonwuelker:
This can be merged now.

PR #385, opened 2026-06-08
Add "aspect-ratio" media feature.

jschwe:
The implementation does the same as what gecko does (except that I didn't add support for `device-aspect-ratio`, hence slightly simpler / inlined.

Servo PR: https://github.com/servo/servo/pull/45490

### DIFF
--- a/style/servo/media_features.rs
+++ b/style/servo/media_features.rs
@@ -7,7 +7,7 @@
 use crate::derives::*;
 use crate::queries::feature::{AllowsRanges, Evaluator, FeatureFlags, QueryFeatureDescription};
 use crate::queries::values::PrefersColorScheme;
-use crate::values::computed::{CSSPixelLength, Context, Resolution};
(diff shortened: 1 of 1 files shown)


Loirooriol (approved):

style/servo/media_features.rs
```diff
@@ -129,8 +129,14 @@ fn eval_any_hover(context: &Context, query_value: Option<Hover>) -> bool {
     eval_hover_capabilities(query_value, context.device().all_pointer_capabilities())
 }
 
+/// https://drafts.csswg.org/mediaqueries-4/#aspect-ratio
```
mrobinson: Raw URLs in rustdoc will break the rustdoc build:

```suggestion
/// <https://drafts.csswg.org/mediaqueries-4/#aspect-ratio>
```

mrobinson (approved):

PR #388, opened 2026-06-12
Remove `overrides_transform_style()` and `get_used_transform_style()`

mrobinson:
It's not possible to properly calculate these only looking at style,
because the results depends on the used `overflow` value.

Servo PR: servo/servo#45629

### DIFF
--- a/style/properties/properties.mako.rs
+++ b/style/properties/properties.mako.rs
@@ -2121,32 +2121,6 @@ impl ComputedValuesInner {
             &position_style.left,
         ))
     }
-
-    /// Return true if the effects force the transform style to be Flat
-    pub fn overrides_transform_style(&self) -> bool {
-        use crate::computed_values::mix_blend_mode::T as MixBlendMode;
-
-        let effects = self.get_effects();
-        // TODO(gw): Add clip-path, isolation, mask-image, mask-border-source when supported.
-        effects.opacity < 1.0 ||
-           !effects.filter.0.is_empty() ||
-           !effects.clip.is_auto() ||
-           effects.mix_blend_mode != MixBlendMode::Normal
-    }
-
-    /// <https://drafts.csswg.org/css-transforms/#grouping-property-values>
-    pub fn get_used_transform_style(&self) -> computed_values::transform_style::T {
-        use crate::computed_values::transform_style::T as TransformStyle;
-
-        let box_ = self.get_box();
-
-        if self.overrides_transform_style() {
-            TransformStyle::Flat
-        } else {
-            // Return the computed value if not overridden by the above exceptions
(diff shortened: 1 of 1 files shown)


style/properties/properties.mako.rs
```diff
@@ -2124,28 +2124,51 @@ impl ComputedValuesInner {
...
+        //    hidden.
+        //
+        // TODO: Support `mask-image`, `mask-border-source`, and `contain`.
         let effects = self.get_effects();
-        // TODO(gw): Add clip-path, isolation, mask-image, mask-border-source when supported.
-        effects.opacity < 1.0 ||
-           !effects.filter.0.is_empty() ||
-           !effects.clip.is_auto() ||
+        !matches!(self.get_box().clone_overflow_x(), Overflow::Visible | Overflow::Clip) ||
+            !matches!(self.get_box().clone_overflow_y(), Overflow::Visible | Overflow::Clip) ||
```
Loirooriol: Should this check the used overflow? E.g. a table row is a transformable element so `transform-style` applies, but `overflow` doesn't. Or propagation of `overflow` to the viewport.

If so, this should move to Servo. May make sense to move it regardless.

mrobinson: Good point. I've modified this PR to simply remove this functionality, have moved it to https://github.com/servo/servo/pull/45629, and ensured that there the used overflow is considered when determining the used `TransformStyle`.

nicoburns:
One thing I've thought about for these kind of values is that it could be useful to pre-compute the bits that can be computed purely from style information and store them on the `ComputedValues`/`ComputedValuesInner` in a bitset (I am imagining that this would be a bit of Servo-only code that runs during the style pass as/just after the computed values for a node are computed). That would make it much cheaper to check rather than doing comparisons on several style properties each time (some of which like the transforms can be quite large).

Not sure how much different it would make to performance, but it'd be an easy optimization to implement, and might also be quite nice in terms of "making the code structure reflect the spec"?

mrobinson:
> Not sure how much different it would make to performance, but it'd be an easy optimization to implement, and might also be quite nice in terms of "making the code structure reflect the spec"?

That's an interesting idea. I think our plan in Servo is to cache these kinds of things on the `Fragment`, but maybe there is room for multiple layers of caching.

style/properties/properties.mako.rs
Loirooriol: Seems you didn't push the removal?

mrobinson: This was the typical situation where I had this in my working directory, but uncommitted. I've pushed it now.

Loirooriol (approved):

PR #391, opened 2026-06-13
Add the pointer/hover media features

### LINKED ISSUE #390: Implement the pointer/hover interaction media features for the servo backend
## Missing: the `pointer` / `hover` interaction media features (servo backend)

Servo's media-feature list (`style/servo/media_features.rs`) doesn't implement the [mediaqueries-4 interaction features](https://drafts.csswg.org/mediaqueries-4/#mf-interaction): `pointer`, `any-pointer`, `hover`, `any-hover`. Since an unknown feature never matches, any rule guarded by one is silently dropped.

### Real-world impact

Modern CSS frameworks emit hover styles behind these queries. Tailwind v4, for example, wraps every `:hover` utility in `@media (hover: hover) { … }`. On a servo-backed engine (e.g. Blitz, which uses Stylo with the servo configuration) this means **all hover styling is dropped** — hover states, button highlights, etc. just don't apply.

### Notes for implementation

- The gecko backend already implements these in `style/gecko/media_features.rs` via a `PointerCapabilities` bitflags concept (COARSE / FINE / HOVER) read off the device.
- The servo `Device` (`style/device/servo.rs`) would need an equivalent capability it can evaluate against — a sensible default for a desktop window is a fine pointer that can hover, ideally overridable for touch-first embedders.

I had a working implementation but it was AI-assisted, so per the [AI-contributions policy](https://book.servo.org/contributing/getting-started.html#ai-contributions) I've closed that PR (#389) and I'm filing this instead so a maintainer can implement it natively. Happy to test any implementation against a real Tailwind-styled app.

nicoburns:
Fixes #390

This feature is quite impactful as it is required for Tailwind's hover styles to work.

Servo PR: https://github.com/servo/servo/pull/45681

### DIFF
--- a/style/device/servo.rs
+++ b/style/device/servo.rs
@@ -13,6 +13,7 @@ use crate::media_queries::MediaType;
 use crate::properties::style_structs::Font;
 use crate::properties::ComputedValues;
 use crate::queries::values::PrefersColorScheme;
+use crate::servo::media_features::PointerCapabilities;
 use crate::values::computed::font::GenericFontFamily;
 use crate::values::computed::{CSSPixelLength, Length, LineHeight, NonNegativeLength};
 use crate::values::specified::color::{ColorSchemeFlags, ForcedColors, SystemColor};
@@ -64,6 +65,12 @@ pub(super) struct ExtraDeviceData {
     /// Whether the user prefers light mode or dark mode
     #[ignore_malloc_size_of = "Pure stack type"]
     prefers_color_scheme: PrefersColorScheme,
+    /// The capabilities of the primary pointer input
+    #[ignore_malloc_size_of = "Pure stack type"]
+    primary_pointer_capabilities: PointerCapabilities,
+    /// The union of the capabilities of all pointer inputs
+    #[ignore_malloc_size_of = "Pure stack type"]
+    all_pointer_capabilities: PointerCapabilities,
     /// An implementation of a trait which implements support for querying font metrics.
     #[ignore_malloc_size_of = "Owned by embedder"]
     font_metrics_provider: Box<dyn FontMetricsProvider>,
@@ -79,6 +86,8 @@ impl Device {
         font_metrics_provider: Box<dyn FontMetricsProvider>,
         default_values: Arc<ComputedValues>,
         prefers_color_scheme: PrefersColorScheme,
+        primary_pointer_capabilities: PointerCapabilities,
+        all_pointer_capabilities: PointerCapabilities,
     ) -> Device {
         let root_style = RwLock::new(Arc::clone(&default_values));
         Device {
@@ -104,6 +113,8 @@ impl Device {
                 device_pixel_ratio,
                 quirks_mode,
                 prefers_color_scheme,
+                primary_pointer_capabilities,
+                all_pointer_capabilities,
                 font_metrics_provider,
             },
         }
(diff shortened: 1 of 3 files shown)


UMCEKO:
Lifesaver, I had some issues on some projects which depends on this package, unblocked thanks to you.

style/servo/media_features.rs
```diff
@@ -60,8 +60,91 @@ fn eval_prefers_color_scheme(context: &Context, query_value: Option<PrefersColor
...
+        const HOVER = 0b100;
+    }
+}
+
+impl Default for PointerCapabilities {
+    #[cfg(any(target_os = "ios", target_os = "android"))]
+    fn default() -> Self {
+        PointerCapabilities::COARSE
+    }
+    #[cfg(not(any(target_os = "ios", target_os = "android")))]
```
jdm: We should probably include ohos here.

jdm (approved):

style/servo/media_features.rs
nicoburns: Added! I did actually try to add `ohos` originally, but got confused because it needs `target_env` rather than `target_os`.

style/servo/media_features.rs
```diff
@@ -46,14 +46,121 @@ fn eval_prefers_color_scheme(context: &Context, query_value: Option<PrefersColor
     }
 }
 
+bitflags! {
```
Loirooriol, line 49: Could have been shared with Gecko

style/servo/media_features.rs
```diff
@@ -46,14 +46,121 @@ fn eval_prefers_color_scheme(context: &Context, query_value: Option<PrefersColor
...
+        /// The input mechanism includes a pointing device of limited accuracy, such as a finger on a touchscreen.
+        const COARSE = 0b001;
+        /// The input mechanism includes an accurate pointing device, such as a mouse.
+        const FINE = 0b010;
+        /// The input mechanism can conveniently hover over elements.
+        const HOVER = 0b100;
+    }
+}
+
+impl Default for PointerCapabilities {
```
Loirooriol, line 62: Not sure about having a random default here, seems like it should be up to Servo to decide to pass the correct thing or lie.

style/device/servo.rs
```diff
@@ -264,6 +275,34 @@ impl Device {
         self.extra.prefers_color_scheme
     }
 
+    /// Set the [`PointerCapbabilities`] value for the primary pointer on this [`Device`]
+    ///
+    /// Note that this does not update any associated `Stylist`. For this you must call
+    /// `Stylist::media_features_change_changed_style` and
+    /// `Stylist::force_stylesheet_origins_dirty`.
+    pub fn set_primary_pointer_capabilities(&mut self, capabilities: PointerCapabilities) {
```
Loirooriol, line 283: Is this used?

nicoburns: No, but the other "device data" properties seem to have getters/setters

style/servo/media_features.rs
nicoburns: It's convenient to have it here rather than in Servo because otherwise it needs to be duplicated into things like tests. But perhaps we could make it an explicitly named method (`PointerCapabilities::guess_from_os()` or similar) rather than using the `Default` trait for this?

style/servo/media_features.rs
nicoburns: Agreed. We should probably do that as a follow-up.

style/servo/media_features.rs
Loirooriol: I think unit tests can just pass an empty bitflags, since they don't have pointer capabilities?

PR #368, opened 2026-05-18
Add device-width/device-height media features.

jdm:
Servo PR: https://github.com/servo/servo/pull/44978

PR #394, opened 2026-06-15
Revert "Add device-width/device-height media features."

jdm:
Reverts servo/stylo#368

More work is required in Servo to integrate this.


nicoburns (approved):

nicoburns:
Yeah, we normally have a policy of "make sure the Servo PR CI green before merging the Stylo PR" for this reason. We should probably write that down somewhere.

PR #400, opened 2026-06-25
Optimize `Animation::iterate()` into `Animation::iterate_by()`

Loirooriol:
This method was performing a single iteration. But most callers want to perform multiple iterations, so they were calling it in a loop.

So this patch replaces `iterate()` with `iterate_by()`, which takes the desired number of iterations as a parameter, and returns the actual number of iterations that were performed.

Servo PR: https://github.com/servo/servo/pull/45978

### DIFF
--- a/style/servo/animation.rs
+++ b/style/servo/animation.rs
@@ -594,30 +594,33 @@ impl Animation {
             return false;
         }
 
-        if self.on_last_iteration() {
-            return false;
-        }
-
-        self.iterate();
-        true
+        self.iterate_by(1.) == 1.
     }
 
-    fn iterate(&mut self) {
-        debug_assert!(!self.on_last_iteration());
+    /// Attempts to advance this animation by `n` iterations, but stops when reaching
+    /// the last iteration, and doesn't perform fractional iterations.
+    /// Returns the actual number of iterations that happened.
+    fn iterate_by(&mut self, n: f64) -> f64 {
+        let n = n.trunc().min(self.remaining_iterations().ceil() - 1.0);
+        if n < 1. {
+            return 0.;
+        }
 
         if let KeyframesIterationState::Finite(ref mut current, max) = self.iteration_state {
-            *current = (*current + 1.).min(max);
+            *current = (*current + n).min(max);
         }
 
         if let AnimationState::Paused(ref mut progress) = self.state {
-            debug_assert!(*progress > 1.);
-            *progress -= 1.;
+            debug_assert!(*progress > n);
+            *progress -= n;
         }
 
         // Update the next iteration direction if applicable.
-        self.started_at += self.duration;
+        self.started_at += self.duration * n;
         match self.direction {
-            AnimationDirection::Alternate | AnimationDirection::AlternateReverse => {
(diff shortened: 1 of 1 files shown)


style/servo/animation.rs
```diff
@@ -594,30 +594,32 @@ impl Animation {
...
-    fn iterate(&mut self) {
-        debug_assert!(!self.on_last_iteration());
+    /// Attempts to advance this animation by `n` iterations, but stops when there are
+    /// no more remaining iterations, and doesn't perform fractional iterations.
+    /// Returns the actual number of iterations that happened.
+    fn iterate_by(&mut self, n: f64) -> f64 {
+        let n = n.min(self.remaining_iterations());
+        if n < 1. {
+            return 0.;
+        }
```
Loirooriol: This is actually a bit different, because when `remaining_iterations() == 1` it's considered to be `on_last_iteration()`, so previously it wasn't iterating, but I don't see why not.

Loirooriol: That broke various tests, so now changing behavior now

yezhizhen (approved):
Nice.

style/servo/animation.rs
```diff
@@ -714,10 +720,7 @@ impl Animation {
             match old_state {
                 Paused(old_progress) => {
                     let mut progress = old_progress + (old_delay - self.delay) / self.duration;
-                    while progress > 1. && !self.on_last_iteration() {
-                        self.iterate();
-                        progress -= 1.;
-                    }
+                    progress -= self.iterate_by(progress);
```
yezhizhen, line 723: The `progress` we pass to `iterate_by` can actually be negative.

The behaviour actually does not change. But was there always a bug here?

style/servo/animation.rs
```diff
@@ -594,30 +594,33 @@ impl Animation {
...
 
-    fn iterate(&mut self) {
-        debug_assert!(!self.on_last_iteration());
+    /// Attempts to advance this animation by `n` iterations, but stops when reaching
+    /// the last iteration, and doesn't perform fractional iterations.
+    /// Returns the actual number of iterations that happened.
+    fn iterate_by(&mut self, n: f64) -> f64 {
+        let n = n.trunc().min(self.remaining_iterations().ceil() - 1.0);
+        if n < 1. {
+            return 0.;
```
yezhizhen, line 606: If `n` passed in is negative, we will return 0. Will that correctly update the state?

mrobinson (approved):

style/servo/animation.rs
Loirooriol: Not sure I'm following, what do you think should happen if `progress` is negative?

style/servo/animation.rs
Loirooriol: In principle yes, the caller was just calling `iterate()` in a loop, and doing `progress -= 1.;` each time. If `n` is negative then the loop condition would have never been true, which is equivalent to subtracting the 0 that is returned here.

style/servo/animation.rs
yezhizhen: Nvm. I think it is correct. We will set the `self.state = Paused(progress);` one line below.

PR #401, opened 2026-06-25
Make `iterate_by()` increment the iteration state for infinite animations

Loirooriol:
It was only incremented for finite animations, but infinite animations need to be incremented too.

Servo PR: https://github.com/servo/servo/pull/45990

### DIFF
--- a/style/servo/animation.rs
+++ b/style/servo/animation.rs
@@ -606,8 +606,13 @@ impl Animation {
             return 0.;
         }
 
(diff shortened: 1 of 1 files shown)


yezhizhen (approved):
Makes sense. But would need refactor after #400.

mrobinson (approved):

Loirooriol:
> But would need refactor after https://github.com/servo/stylo/pull/400

Yes, just a matter of adding `n` instead of `1.`

PR #405, opened 2026-06-28
Use static slice for supported media features.

jdm:
This reduces the risk of merge conflicts when adding new media features.


Loirooriol (commented):
Interesting. I guess while you are at it, could do it for Gecko's `MEDIA_FEATURES`. And for `CONTAINER_FEATURES`. And `FeatureType::features()` does no longer need to use `&` for them.

PR #411, opened 2026-07-02
Implement the `Device::is_dark_color_scheme` function

nicoburns:
This is necessary (but may not be sufficient) to make `light-dark()` functions work.

Implementation based on the Gecko implementation: https://searchfox.org/firefox-main/source/widget/nsXPLookAndFeel.cpp#1296

The logic is:
- If the user specifies a single [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/color-scheme) (`light` or `dark`) value then use that (value comes in from the `ColorSchemeFlags`)
- Else fall back to preferred color scheme from the `Device`


style/device/servo.rs
```diff
@@ -321,9 +321,24 @@ impl Device {
     pub fn all_pointer_capabilities(&self) -> PointerCapabilities {
         self.extra.all_pointer_capabilities
     }
-
-    pub(crate) fn is_dark_color_scheme(&self, _: ColorSchemeFlags) -> bool {
-        false
+    pub(crate) fn is_dark_color_scheme(&self, flags: ColorSchemeFlags) -> bool {
+        // The below is based on the implementation from Gecko:
+        // https://searchfox.org/firefox-main/source/widget/nsXPLookAndFeel.cpp#1296
```
Loirooriol: Line number is useless if it's not a permalink.

https://searchfox.org/firefox-main/rev/0a7f146ccac85b8f413264042dcd764028d419ec/widget/nsXPLookAndFeel.cpp#1296

Loirooriol (commented):
I'm not sure I see the point of adding a function that isn't used. If you are implementing `light-dark()`, wouldn't it be better to add it as part of that PR?

Loirooriol:
Sorry, somehow failed to see the function already exists as returning false.

Loirooriol (approved):
So I guess this looks good if you use a permalink.

It's not really changing behavior because `color-scheme` doesn't parse, right?

nicoburns:
> It's not really changing behavior because color-scheme doesn't parse, right?

In theory it should make it respect the system theme (light/dark mode) which we already plug in to Stylo for `@prefer-color-scheme`). But I'm not sure if there are other places we need to change to make this work properly. `color-scheme` is just a per-element override for that.

Possibly we ought to enable `color-scheme` parsing, as I think most of implementation of that might actually be within Stylo?

Loirooriol:
> In theory it should make it respect the system theme

Then please ensure test coverage

nicoburns:
Closing in favour of #438

PR #412, opened 2026-07-02
Group media data into a struct + add getters/setters for the entire struct

nicoburns:
This is aimed at trying to reduce the proliferation of getters/setters for individual media data properties by:

- Grouping them into a struct
- Adding getters/setters for the entire struct.

It also (IMO) makes the `Device` constructor a lot nicer as you get field names rather than having to try to match up function parameter order.

In future we could consider removing the individual getters/setters (or making them private where they are called internally by Stylo), but this PR does not yet do that.

This PR is implemented on top of:
- https://github.com/servo/stylo/pull/410


jdm (approved):
Agreed; I think the current code is only going to expand over time so this is an improvement.

PR #422, opened 2026-07-09
Also update `AnimationState::Paused` in `Animation::update_from_other` if delay changes

mrobinson:
If the delay in an animation changes, we still might need to update the
`Paused` state if the animation is moving from `Playing` to `Paused`.

Servo PR: servo/servo#46689

### DIFF
--- a/style/servo/animation.rs
+++ b/style/servo/animation.rs
@@ -737,15 +737,17 @@ impl Animation {
                         self.state = Pending;
                     }
                 },
-                _ => {
-                    // Running or Pending — re-advance iterations from a fresh
-                    // iteration state.
-                    let starting_progress = (now - self.started_at) / self.duration;
+                Canceled | Pending | Running => {
+                    // Re-advance iterations from a fresh iteration state.
+                    let new_starting_progress = (now - self.started_at) / self.duration;
                     match self.iteration_state {
                         KeyframesIterationState::Finite(ref mut current, _) => *current = 0.0,
                         _ => {},
                     }
-                    self.iterate_by(starting_progress);
+                    if let AnimationState::Paused(ref mut starting_progress) = &mut self.state {
+                        *starting_progress = new_starting_progress;
+                    }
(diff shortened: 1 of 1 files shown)


mrobinson:
Servo PR for this incoming.

mrobinson:
This is ready for review cc @yezhizhen .

style/servo/animation.rs
yezhizhen, line 726: Paused is here

style/servo/animation.rs
```diff
@@ -738,14 +738,17 @@ impl Animation {
                     }
                 },
                 _ => {
-                    // Running or Pending — re-advance iterations from a fresh
+                    // Running, Pending, or Paused — re-advance iterations from a fresh
```
yezhizhen:
```suggestion
                    // Running or Pending — re-advance iterations from a fresh
```
This can't be paused. We are talking about old state.

style/servo/animation.rs
```diff
@@ -738,14 +738,17 @@ impl Animation {
...
                     // iteration state.
-                    let starting_progress = (now - self.started_at) / self.duration;
+                    let new_starting_progress = (now - self.started_at) / self.duration;
                     match self.iteration_state {
                         KeyframesIterationState::Finite(ref mut current, _) => *current = 0.0,
                         _ => {},
                     }
-                    self.iterate_by(starting_progress);
+                    if let AnimationState::Paused(ref mut starting_progress) = &mut self.state {
```
yezhizhen, line 747: Here we are talking about new state, which contain the case going from Playing to Paused.

yezhizhen (approved):
Comment isn't accurate

style/servo/animation.rs
mrobinson: Hrm. Good point. The original comment neglected to include `Canceled` here, so I've converted this code to use an explicit match instead of a wildcard, which is slightly more self-documenting.

mrobinson:
@yezhizhen Thanks for the review.

PR #424, opened 2026-07-11
Stop requiring CI jobs to pass on Mac

Loirooriol:
This is a partial revert of #340. Waiting for an available macOS runner frequently takes very long, which makes it annoying when trying to land multiple patches, since they need to also be coordinated with the companion patches in Servo.

So I think it's better to not wait for the macOS job after all. We will still wait for the Linux and Windows ones.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,8 +65,9 @@ jobs:
     needs:
       - linux-debug
       - linux-release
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
Yeah, I thought that might cause issues. Hopefully we get more macOS CI capacity at some point!

PR #425, opened 2026-07-12
Make `text-decoration-color` only generate repaint damage

nicoburns:
This probably isn't a property that changes frequently in practice, but it seems to me that this ought to only generate `repaint` damage.

### DIFF
--- a/style/properties/longhands.toml
+++ b/style/properties/longhands.toml
@@ -2767,7 +2767,7 @@ initial = "computed_value::T::currentcolor()"
 struct = "text"
 ignored_when_colors_disabled = true
 spec = "https://drafts.csswg.org/css-text-decor/#propdef-text-decoration-color"
-servo_restyle_damage = "recalculate_overflow"
+servo_restyle_damage = "repaint"
 affects = "paint"
 
 [text-decoration-inset]


style/properties/shorthands.rs
```diff
@@ -3246,7 +3246,6 @@ pub mod text_decoration {
     pub use crate::properties::generated::shorthands::text_decoration::*;
 
     use super::*;
-    #[cfg(feature = "gecko")]
```
Loirooriol, line 3249: Why are you removing this?

nicoburns: Ack, that got committed to the wrong branch. Removed.

mrobinson:
Text decorations are collected during stacking context tree construction in Servo, so they'll need stacking context tree damage.

mrobinson (changes requested):
Please see my comment above.

nicoburns:
I see. I guess I'll close this then.

Loirooriol:
I guess it could use `rebuild_stacking_context`.
But I'm seeing that e.g. `isolation` has `servo_restyle_damage = "repaint"` even though it can establish a stacking context... So not sure if it's broken, or we aren't correctly handling damages?

PR #426, opened 2026-07-13
Enable text-decoration-thickness; more text decoration properties for Blitz

nicoburns:
Enables `text-decoration-thickness`, `text-decoration-inset`, `text-underline-offset`, and `text-underline-position` for Servo builds of Stylo, behind the `layout.unimplemented` feature flag.

The shorthand `text-decoration` is updated to account for `text-decoration-thickness` being enabled (this aligns Servo's implementation with Gecko's for this property).

Servo PR:

- https://github.com/servo/servo/pull/46592


style/properties/shorthands.rs
```diff
@@ -3305,19 +3295,19 @@ pub mod text_decoration {
             let is_solid_style =
                 *self.text_decoration_style == text_decoration_style::SpecifiedValue::Solid;
             let is_current_color = *self.text_decoration_color == Color::CurrentColor;
-            #[cfg(feature = "gecko")]
-            let is_auto_thickness = self.text_decoration_thickness.is_auto();
-            #[cfg(feature = "servo")]
-            let is_auto_thickness = true;
+            let is_auto_thickness = self
+                .text_decoration_thickness
+                .is_none_or(|thickness| thickness.is_auto());
```
Loirooriol: This is not going to compile for Gecko

Loirooriol (changes requested):

style/properties/shorthands.rs
nicoburns: Ah, the optionality is per-engine? Makes sense! There is now a servo-only `unwrap_or`.

style/properties/longhands.toml
```diff
@@ -2645,17 +2645,19 @@ affects = "overflow"
 type = "LengthPercentageOrAuto"
 initial = "computed::LengthPercentageOrAuto::auto()"
 struct = "inherited_text"
-engine = "gecko"
+servo_pref = "layout.unimplemented"
 spec = "https://drafts.csswg.org/css-text-decor-4/#underline-offset"
+servo_restyle_damage = "recalculate_overflow"
```
Loirooriol: Seems weird to have this for an unimplemented property, does Blitz have the same damage system?

Also it's more than needed, since text decorations are considered ink overflow, so no need to recalculate scrollable regions.

nicoburns: > Seems weird to have this for an unimplemented property

Well otherwise it's implicitly defaulting to `rebuild_box`. Better to have something accurate, no?

> does Blitz have the same damage system?

Yes. I don't think there's any other way to get damage out of Stylo. I did put up https://github.com/servo/stylo/pull/224 which would give control to the embedder. But in practice I haven't really run into problems with just using Servo's setup, so I haven't pushed that.

> Also it's more than needed, since text decorations are considered ink overflow, so no need to recalculate scrollable regions.

Ah, based on some of the other properties that have `recalculate_overflow` set, I thought that it must be required for ink overflow. Can definitely change it to `repaint` if that's all that's needed.

nicoburns: I've changed the damage to repaint

style/properties/shorthands.rs
```diff
@@ -3239,20 +3239,30 @@ pub mod text_decoration {
...
+    #[cfg(feature = "gecko")]
+    #[inline(always)]
+    fn parse_thickness() -> bool {
+        true
+    }
+
+    #[cfg(feature = "servo")]
+    #[inline(always)]
+    fn parse_thickness() -> bool {
+        static_prefs::pref!("layout.unimplemented")
```
Loirooriol: At this point, I think you should add a specific pref for this. Gecko used to have `layout.css.text-decoration-thickness.enabled`. https://github.com/mozilla-firefox/firefox/commit/f4ee21ca1bb3496cbb5b4df78cfe696e53d99c69#diff-fbb611c8fcb79e0ee8911d8e9ce8cbdc8894e04330d01b0f84500c5fcf55b831

nicoburns: @Loirooriol I've gone ahead and implemented `text-decoration-thickness` in Servo instead (so we can remove the pref here). All conditional "shorthand" code has been removed from this PR.

style/properties/longhands.toml
Loirooriol: As stated in https://github.com/servo/stylo/pull/425#issuecomment-4958592511, it should be `rebuild_stacking_context`. With just `repaint`, this won't get updated:

```html
<!DOCTYPE html>
<div style="text-decoration: underline">
  <div style="isolation: isolate">lorem ipsum</div>
</div>
<script>
requestAnimationFrame(() => requestAnimationFrame(() => {
  document.querySelector("div").style.textDecorationThickness = "10px";
}));
</script>
```

nicoburns: I have now changed the damage to `rebuild_stacking_context`. Definitely seems like a general review of damage for all properties might be a good idea!

style/properties/shorthands.rs
```diff
@@ -3292,23 +3282,21 @@ pub mod text_decoration {
         where
             W: fmt::Write,
         {
+            use crate::values::generics::text::GenericTextDecorationLength;
```
Loirooriol: Seems unnecessary

style/properties/shorthands.rs
```diff
@@ -3292,23 +3282,21 @@ pub mod text_decoration {
...
             use crate::values::specified::Color;
             use crate::values::specified::TextDecorationLine;
 
             let is_solid_style =
                 *self.text_decoration_style == text_decoration_style::SpecifiedValue::Solid;
             let is_current_color = *self.text_decoration_color == Color::CurrentColor;
-            #[cfg(feature = "gecko")]
+
```
Loirooriol: No need for newline

Loirooriol (approved):
Looks good with nits

PR #431, opened 2026-07-27
Avoid panic in `parse_legacy_color` on whitespace-only input

nicoburns:
## Bug

`parse_legacy_color` only checked emptiness *before* stripping HTML whitespace, so a value like `<body bgcolor=" ">` reached step 9's `input.as_bytes()[0]` with an empty string and panicked with an index-out-of-bounds, a crash triggerable from untrusted HTML.

## Changes made

Two changes made in `style/servo/attr.rs`:

1. Check for empty input again after stripping whitespace.
2. Use the built-in stdlib `strip_prefixes` function which is tolerant of empty input. Technically not necessarily given fix, but it's a nice readability improvement.

This essentially fixes the bug twice, but I think a little defence in depth is nice here

---

AI declaration: this bug was found by an AI-security scan, and fix included here was proposed by the AI.
(it has been verified to be a genuine bug and sensible fix by me).


mrobinson:
It would be better to submit this upstream in Gecko, I think. As far as I know this is shared code.

nicoburns:
It's in a dir called "servo", so I think it's probably not shared. But I can probably submit upstream anyway.

mrobinson:
> It's in a dir called "servo", so I think it's probably not shared. But I can probably submit upstream anyway.

Ah, sorry. I missed that. It's not shared, so submitting it here makes sense.

mrobinson:
@nicoburns
 1. Is there a test for this change? 
 2. I don't think you need an AI Declaration for this change. If you wrote it yourself, it doesn't matter if a LLM identified the original bug: https://book.servo.org/contributing/getting-started.html#ai-policy-faq

nicoburns:
I have added a test, although `cargo test` does not compile with Stylo `main` for other reasons. I have not included fixes for that as it seemed out of scope (but I did locally patch so I could run the included test).

style/servo/attr.rs
```diff
@@ -528,6 +528,9 @@ pub fn parse_legacy_color(mut input: &str) -> Result<AbsoluteColor, ()> {
 
     // Step 3.
     input = input.trim_matches(HTML_SPACE_CHARACTERS);
+    if input.is_empty() {
```
Loirooriol, line 531: This seems a spec problem https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-a-legacy-colour-value

Please file an issue.

style/servo/attr.rs
```diff
@@ -712,3 +715,12 @@ pub struct AttrIdentifier {
...
 }
+
+#[cfg(test)]
+mod test {
+    use crate::attr::parse_legacy_color;
+    #[test]
+    fn parsing_whitespace_only_color_does_not_panic() {
+        assert!(parse_legacy_color(" ").is_err());
+    }
+}
```
Loirooriol, line 726: This isn't checked, so please make sure there is WPT coverage

PR #432, opened 2026-08-03
Remove `type Impl` from traits PseudoElement and NonTSPseudoClass

dislogical:
These aren't actually used anywhere, and make re-using Pseudo implementations between different parsers MUCH more complicated.

This is technically a breaking change, as downstreams will need to remove the `type` definition, but that's all that's required (see style/servo/selector_parser.rs for required changes).

### DIFF
--- a/selectors/parser.rs
+++ b/selectors/parser.rs
@@ -33,9 +33,6 @@ use to_shmem_derive::ToShmem;
 
 /// A trait that represents a pseudo-element.
 pub trait PseudoElement: Sized + ToCss {
-    /// The `SelectorImpl` this pseudo-element is used for.
-    type Impl: SelectorImpl;
-
     /// Whether the pseudo-element supports a given state selector to the right
     /// of it.
     fn accepts_state_pseudo_classes(&self) -> bool {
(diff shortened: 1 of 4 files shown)


mrobinson:
Are these used in Gecko? If so, I'm not sure we should remove them. We are downstream of Gecko, so this sort of change should likely happen there.

dislogical:
Great question! I’m new here and can’t find any obvious docs on how to check, is there anything I should be looking at?

nicoburns:
Try applying your patch to https://github.com/mozilla-firefox/firefox (stylo is in the servo directory) and see if it builds.

dislogical:
Ok just confirmed that this patch _does_ compile in upstream Firefox. Does that mean I should submit this patch there?

mrobinson:
@dislogical That would be ideal as then we wouldn't have to upstream it and would just incorporate it the next time we sync stylo.

dislogical:
@mrobinson sounds good, I'll give it a shot! I know their barrier to contribution entry is a bit higher than "PRs welcome," so I'm hoping they'll take such a small change.

dislogical:
Upstream review: https://phabricator.services.mozilla.com/D316430

mrobinson:
@dislogical Thank you so much! I'll close this in favor of the upstream contribution.

dislogical:
No problem! It was easier than I thought.

How often do y'all pull updates from upstream?

Loirooriol:
Every month

PR #434, opened 2026-08-04
enable Pseudo Element ::-moz-progress-bar

rayguo17:
Enable PseudoElement MozProgressBar for use in Servo
Companion Servo PR: https://github.com/servo/servo/pull/46974


style/servo/selector_parser.rs
```diff
@@ -69,6 +69,7 @@ pub enum PseudoElement {
     SliderFill,
     SliderThumb,
     SliderTrack,
+    MozProgressBar,
```
Loirooriol, line 72: I guess the reason for using the Moz prefix (instead of Servo) is because Gecko exposes it to authors?

rayguo17: Yes.

Loirooriol (approved):

PR #437, opened 2026-08-04
Change repr of Servo enum PseudoElement from usize to u8

SimonSapin:
Every fragment in the fragment tree stores:

```rust
pub struct PseudoElementChain {
    pub primary: Option<PseudoElement>,
    pub secondary: Option<PseudoElement>,
}
```

… so this change reduces memory use slightly


mrobinson (approved):

mrobinson:
Normally we link to a Servo PR with changes, though it seems quite unlikely that this will break Servo.

Loirooriol:
Gecko also uses u8: https://searchfox.org/firefox-main/rev/28b5e86da948982342836c3cca50049659f0dc70/servo/components/style/gecko/pseudo_element_definition.mako.rs#14

PR #438, opened 2026-08-05
Support color-scheme in Servo

TimvdLippe:
With these changes, both the `color-scheme` CSS property is supported, as well as the required logic to act upon it.

Supersedes and closes #411

Servo PR: https://github.com/servo/servo/pull/47042/

### DIFF
--- a/style/device/servo.rs
+++ b/style/device/servo.rs
@@ -333,8 +333,21 @@ impl Device {
         self.extra.all_pointer_capabilities
     }
 
-    pub(crate) fn is_dark_color_scheme(&self, _: ColorSchemeFlags) -> bool {
-        false
+    pub(crate) fn is_dark_color_scheme(&self, color_scheme_flags: ColorSchemeFlags) -> bool {
+        // Inspired by
+        // https://searchfox.org/firefox-main/rev/0a7f146ccac85b8f413264042dcd764028d419ec/widget/nsXPLookAndFeel.cpp#1296
+        let prefers_color_scheme = self.color_scheme();
+        let supports_dark_mode = color_scheme_flags.contains(ColorSchemeFlags::DARK);
+        let supports_light_mode = color_scheme_flags.contains(ColorSchemeFlags::LIGHT);
+
+        // If only one is supported, then use dark mode if it was the supported one.
+        if supports_dark_mode != supports_light_mode {
+            return supports_dark_mode;
+        }
+
+        // If either both or none are supported, then use the preferred color scheme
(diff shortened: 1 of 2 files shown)


nicoburns (approved):

mrobinson (approved):

mrobinson:
Is there a corresponding Servo PR for this change? Typically for Stylo PRs they should have a link to a Servo PR unless it is guaranteed to not change test results / affect the build.

TimvdLippe:
It's still a draft as I ran out of time to create the Servo PR. Once I have that finished, I will mark this PR as ready

TimvdLippe:
Stylo is currently being upgraded in https://github.com/servo/servo/pull/46987 which means that test results will be muddied by the upgrade. Will wait on that

style/device/servo.rs
```diff
@@ -333,8 +333,27 @@ impl Device {
...
+        }
+
+        // If only one is chosen, then it should only support dark mode if it was dark
+        // mode enabled.
+        if supports_dark_mode || supports_light_mode {
+            return supports_dark_mode;
+        }
+
+        // Neither was chosen. In that case, fallback to the preferred color scheme
+        return prefers_color_scheme == PrefersColorScheme::Dark;
```
Loirooriol: I think this would be simpler as

```suggestion
        // If only one is supported, then use dark mode if it was the supported one.
        if supports_dark_mode != supports_light_mode {
            return supports_dark_mode;
        }

        // If either both or none are supported, then use the preferred color scheme
        // to determine whether the user wants dark mode.
        return prefers_color_scheme == PrefersColorScheme::Dark;
```

TimvdLippe:
Keeping this as draft until the Stylo upgrade has completed

nicoburns:
I'm interested in what Servo-side changes will be required for this. My tentative conclusion when I looked into this was that it might actually work entirely within Stylo (aside from the already existing theme support on the Device).

TimvdLippe:
These were the required changes: https://github.com/servo/servo/pull/47042 (at least the ones I figured out up to this point)

nicoburns:
Ah ok, just CSS and test expectation changes. Makes sense.

Loirooriol:
> Keeping this as draft until the Stylo upgrade has completed

No need to wait, I can just cherry-pick it if it lands first

TimvdLippe:
Okay cool. In that case, both PRs are ready for review. But I think everybody basically approves already 😂

mrobinson (approved):

PR #440, opened 2026-08-10
Derive `Clone` on SourceSize and SourceSizeList

lumiscosity:
All members of these structs already derive/implement Clone, so this is trivial. Required for cloning an HTMLLinkElement's source set when constructing LinkProcessingOptions.


jdm (approved):

jdm:
We should submit this upstream as well, since it's a type that's shared with Gecko.

jdm:
https://phabricator.services.mozilla.com/D317614

PR #441, opened 2026-08-10
Model the CSS property database in Rust

pathscale:
First step of removing Python from property generation, and the one that is unblocked regardless of what replaces Mako.

`data.py` builds the property model from the TOML files next to it and hands it to Mako, so reading a property meant running a vendored Python interpreter. This crate is that model as types: 429 longhands, 92 shorthands and the three at-rule descriptor sets, parsed by serde from the same files, with no interpreter and no dependency on the `style` crate.

Nothing in the build changes. `style/build.rs` is untouched, the committed `properties/generated/servo` artifacts are untouched, and `build.py` still runs exactly as before. This adds a crate and nothing consumes it yet.

## Why `deny_unknown_fields` is the point

`data.py` reads these files with attribute lookups and defaults, so a misspelled key is silently ignored. `servo_pre` instead of `servo_pref` yields a property that quietly ships with no preference gate and no error anywhere. Every struct here rejects unknown fields, which turns that into a parse failure naming the file, the property and the key. Building the model was itself the test: it found `none_value`, `gecko_aliases` and the descriptors' `parser` key, none of which are in the schema documentation.

## Two warts kept behind accessors

`flags` is a space separated string, and a keyword's `gecko_aliases` are `"alias=target"` pairs. Both are structure encoded inside strings. The fields are private and the accessors split them, so the generator never parses either again.

`initial` stays a string, because it is a Rust expression rather than data. No serialisation format makes an embedded expression type-safe.

## The schema comment had drifted

The 52-line header in `longhands.toml` still describes `keyword.values` as a space separated list where the file uses an array. The doc comments on these structs are that schema, kept where it cannot drift from the parser. The header should be deleted in a follow-up.

## Verification

`cargo test -p stylo_property_model`: 6 pass. `cargo fmt --check` and `cargo clippy --all-targets` are clean.

The tests pin what the files contain rather than what they might: the counts, that every shorthand names a longhand that exists, that engine filtering excludes the other engine's properties, and the flag vocabulary, so a new flag arriving from upstream fails a test instead of becoming a string nothing reads.

## What this does not decide

Whether the templates become MiniJinja or the generator emits Rust directly. That turns on whether you keep rebasing on mozilla-central, and this model is needed either way.


pathscale:
Opened against the wrong repository by mistake. This belongs on the pathscale/stylo-less-py fork, not upstream. Apologies for the noise.

PR #442, opened 2026-08-13
Enable `text-transform: math-auto` for Servo

SimonSapin:
Drive-by: `math-auto` is not a case transform. It is more like `none`: a stand-alone value that cannot be combined with `full-width` or `full-size-kana` in spec grammar.

Servo PR: https://github.com/servo/servo/pull/47225

### DIFF
--- a/style/values/specified/text.rs
+++ b/style/values/specified/text.rs
@@ -372,9 +372,6 @@ pub enum TextTransformCase {
     Lowercase,
     /// Capitalize each word.
     Capitalize,
-    /// Automatic italicization of math variables.
-    #[cfg(feature = "gecko")]
-    MathAuto,
 }
 
 #[derive(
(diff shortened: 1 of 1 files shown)


style/values/specified/text.rs
```diff
@@ -372,9 +372,6 @@ pub enum TextTransformCase {
     Lowercase,
     /// Capitalize each word.
     Capitalize,
-    /// Automatic italicization of math variables.
-    #[cfg(feature = "gecko")]
-    MathAuto,
```
SimonSapin, line 377: This won’t affect Gecko since it doesn’t use this enum at all: https://searchfox.org/firefox-main/search?q=TextTransformCase

mrobinson (approved):

style/values/specified/text.rs
```diff
@@ -425,17 +411,13 @@ bitflags! {
         const LOWERCASE = 1 << 1;
         /// Capitalize each word.
         const CAPITALIZE = 1 << 2;
-        /// Automatic italicization of math variables.
-        #[cfg(feature = "gecko")]
-        const MATH_AUTO = 1 << 3;
 
-        /// All the case transforms, which are exclusive with each other.
-        #[cfg(feature = "gecko")]
-        const CASE_TRANSFORMS = Self::UPPERCASE.0 | Self::LOWERCASE.0 | Self::CAPITALIZE.0 | Self::MATH_AUTO.0;
```
Loirooriol, line 434: Note Gecko expects `CASE_TRANSFORMS` to include `MATH_AUTO`: https://searchfox.org/firefox-main/rev/c6e08963c19ca94be777dd97bd61ad9d96d4c2f2/layout/generic/nsTextRunTransformations.cpp#389,756

So I'll undo this part when exporting to Firefox.

SimonSapin: Thanks for taking care of this. I looked for Gecko uses of the `TextTransformCase` enum but missed looking for their use of the `CASE_TRANSFORMS` const

Loirooriol: Yes, I have also sent a patch to change that Gecko code to use `TextTransformCase`, since that seems more robust.

PR #444, opened 2026-08-14
Adjust restyle damage for clip-path.

jdm:
clip-path's calculations in servo occur while building the stacking context. If the value changes, we need to rebuild the stacking context.

https://github.com/servo/servo/pull/47256


mrobinson (approved):
This makes sense. Clipping is calculated during stacking context construction.

PR #445, opened 2026-08-20
Implement `animation-composition` for Servo

nicoburns:
Changes made:
- Implement the [animation-composition](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation-composition) CSS property.
- Shares animation matrix decomposition logic with Gecko (fixes a couple of regressions due to the Servo version of the code not rejecting matrices with a zero determinant)

Implementation is entirely in Stylo.

Servo PR: https://github.com/servo/servo/pull/47405

### DIFF
--- a/style/values/animated/transform.rs
+++ b/style/values/animated/transform.rs
@@ -249,18 +249,6 @@ impl From<MatrixDecomposed2D> for Matrix3D {
 }
 
 impl Animate for Matrix {
-    #[cfg(feature = "servo")]
-    fn animate(&self, other: &Self, procedure: Procedure) -> Result<Self, ()> {
-        let this = Matrix3D::from(*self);
-        let other = Matrix3D::from(*other);
-        let this = MatrixDecomposed2D::from(this);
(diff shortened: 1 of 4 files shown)


mrobinson (approved):

style/values/animated/transform.rs
```diff
@@ -249,18 +249,6 @@ impl From<MatrixDecomposed2D> for Matrix3D {
...
-        let this = Matrix3D::from(*self);
-        let other = Matrix3D::from(*other);
-        let this = MatrixDecomposed2D::from(this);
-        let other = MatrixDecomposed2D::from(other);
-        Matrix3D::from(this.animate(&other, procedure)?).into_2d()
-    }
-
-    #[cfg(feature = "gecko")]
-    // Gecko doesn't exactly follow the spec here; we use a different procedure
-    // to match it
```
Loirooriol, line 263: Why are you removing these comments? Was Gecko actually matching the spec here?

PR #451, opened 2026-09-01
Add atoms for cue events

TimvdLippe:
These events are fired for text track cues in track elements.

Part of servo/servo#46288

### DIFF
--- a/stylo_atoms/static_atoms.txt
+++ b/stylo_atoms/static_atoms.txt
@@ -45,6 +45,7 @@ compositionend
 compositionstart
 compositionupdate
 controllerchange
+cuechange
 cursive
 dark
 datachannel
@@ -59,7 +60,9 @@ email
 emptied
 end
 ended
+enter
 error
+exit
 fantasy
 fetch
 file


jdm:
I think you want https://github.com/servo/html5ever/blob/main/web_atoms/local_names.txt instead?

TimvdLippe:
That's what I originally thought and did, and then it didn't compile. No, `fire_event` takes a `style::Atom`: https://github.com/servo/servo/blob/d8098b69ff79dc59971fa18d2edeaecb415830dc/components/script/dom/event/eventtarget.rs#L810 and https://github.com/servo/servo/blob/d8098b69ff79dc59971fa18d2edeaecb415830dc/components/script/dom/event/eventtarget.rs#L31

jdm:
Oh, huh. That's probably something we should fix.

TimvdLippe:
@jdm would you be okay to land this PR and then the Servo PR and afterwards fix those atom usages? It feels like that's a bigger undertaking and atoms have been a bit off a mess for a while. I would like to continue my work on tracks now that I have momentum again. If you think we should do it, that's also fine, but then it will take a while.

jdm:
Oh, yeah, I didn't mean to derail any work here

jdm (approved):

PR #460, opened 2026-09-10
Upgrade to rust-phf 0.14 and intermediate dependencies

SimonSapin:


jdm:
web_atoms has now been published, along with the rest of the html5ever crates.

style/bloom.rs
```diff
@@ -401,3 +401,14 @@ impl<E: TElement> StyleBloom<E> {
...
+pub(crate) trait AtomExt {
+    fn get_hash32(&self) -> u32;
+}
+
+impl<Static: string_cache::StaticAtomSet> AtomExt for string_cache::Atom<Static> {
+    fn get_hash32(&self) -> u32 {
+        let hash64 = self.get_hash();
+        (hash64 >> 32) as u32 ^ (hash64 as u32)
+    }
+}
```
SimonSapin, line 414: I believe this preserves the previous behavior

@emilio Could the bloom filter take advantage of a 64-bit hash?

emilio: Potentially, but at the cost of more memory usage. For the ancestor filter seems fine since that's temporary, for the smaller filters we use it's a bit more unfortunate...

SimonSapin: I see, thanks. If that’s not an improvement let’s keep it as-is then

Cargo.toml
```diff
@@ -34,3 +34,8 @@ dom             = { version = "0.21.0", path = "./stylo_dom", package = "stylo_d
 style_traits    = { version = "0.21.0", path = "./style_traits", features = ["servo"], package = "stylo_traits"}
 style_derive    = { version = "0.21.0", path = "./style_derive", package = "stylo_derive"}
 stylo           = { version = "0.21.0", path = "./style" }
+
+# external dependencies
+cssparser = "0.38"
+string_cache = { version = "0.11" }
+string_cache_codegen = { version = "0.11" }
```
nicoburns: Using workspace dependencies is a bit awkward for Stylo, because (IIRC) the workspace `Cargo.toml` is not synced between Gecko and Servo repos. Can you turn these back into regular dependencies?

nicoburns (changes requested):
See note on workspace deps. Otherwise looks good to me.

Cargo.toml
SimonSapin: Oh I didn’t realize. Reverted

nicoburns (approved):

Loirooriol:
The top comment appears in the commit message, so I have removed the obsolete

> ~~Blocked on a new release of web_atoms~~

SimonSapin:
Removing from queue because servo with this Stylo version panics with `[2026-09-14T17:33:05Z ERROR servoshell::panic_hook] internal error: entered unreachable code: Called into PrecomputedHasher with something that isn't a u32`

SimonSapin:
I’ve pushed a commit that fixed that panic, please take another look

nicoburns (approved):

SimonSapin:
Let’s investigate [WPT regressions](https://github.com/servo/servo/pull/48080#issuecomment-5669685965) before merging this

PR #461, opened 2026-09-12
Reland "Bug 2064900 - Always borrow applicable declarations"

Loirooriol:
It was reverted during the last sync, because adapting Servo to it wasn't straightforward.

For context, see https://phabricator.services.mozilla.com/D322671 and https://bugzil.la/2064900


mrobinson (approved):
Assuming this is just re-landing upstream code, but if there are changes, it would be great to highlight them.

Loirooriol:
There is no change, the tricky part will be in the Servo repo

PR #462, opened 2026-09-17
Enable `user-select` for Servo + `contain` value

SimonSapin:
Servo PR: https://github.com/servo/servo/pull/48137


mrobinson (approved):

jdm:
This shouldn't have been merged before the Servo PR had an approving merge.

mrobinson:
Apologies, the Servo PR is almost ready to land and I'll do that later this morning.

SimonSapin:
My bad, I waited for the stylo PR to be merged in order to have its definitive git commit hash before taking the servo PR out of draft and asking for its review, but I see how it should be the other way around to avoid drift. I’ll keep it in mind for future PRs. (Should we have `CONTRIBUTING.md` or something to explain this?)

(Although in this case, if it comes to that Servo can be updated to a newer Stylo with this change and will compile. At worst it’ll temporarily parse a property without implementing it.)

PR #463, opened 2026-09-21
Invalidate animation cascade rules on cancel, keyframes removal, and display:none

### LINKED ISSUE #48225: Canceling a @keyframes animation freezes the computed style for a later transition on the same element
**Describe the bug:**
When canceling a `@keyframes` animation and subsequently initiating a CSS transition on the same element, the computed `transform` may remain static and fail to update. This occurs despite the transition being correctly declared and the `animation-name` set to `none`. However, applying the same transition to a new element functions as expected.

---

**To Reproduce:**
```html
<!doctype html>
<meta charset="utf-8" />
<style>
#box {
	position: absolute;
	top: 100px;
	left: 24px;
	width: 64px;
	height: 64px;
	background: rgb(30, 90, 210);
}
</style>
<div id="box"></div> <pre id="log"></pre>
<script>
const box = document.getElementById('box');
const log = document.getElementById('log');
const sleep = ms => new Promise(r => setTimeout(r, ms));
const kf = document.createElement('style');
kf.textContent = '@keyframes moving {from {transform:translateX(0px);opacity:.2;width:64px;background-color:rgb(30,90,210)} to {transform:translateX(400px);opacity:1;width:140px;background-color:rgb(230,50,70)}}';
document.head.append(kf);
let running = true;
(function tick() {
	if(!running) return;
	log.textContent = 'transform: ' + getComputedStyle(box).transform + '\nanimationName: ' + getComputedStyle(box).animationName;
	requestAnimationFrame(tick);
})();
(async() => {
	box.style.animation = 'moving 1000ms linear infinite alternate';
	await sleep(2300);
	kf.remove();
	await sleep(50);
	box.style.cssText = '';
	await sleep(200);
	box.style.transform = 'translateX(0px)';
	void box.offsetWidth;
	box.style.transition = 'transform 1000ms linear';
	for(let n = 0; n < 4; n++) {
		box.style.transform = n % 2 ? 'translateX(0px)' : 'translateX(400px)';
		await sleep(1100);
	}
})();
</script>
```

After cancellation, `#box` remains frozen at its last animated `transform` value instead of transitioning between 0 and 400 px, even though `animation-name` is `none`.

The cause is a missing dirty flag update in Stylo’s `ElementAnimationSet::update_animations_for_new_style` ([`style/servo/animation.rs:1288-1304`](https://github.com/servo/stylo/blob/1581aae4050b13112c1a2cc108b3d42619992297/style/servo/animation.rs#L1288-L1304)).. This path sets the animation state to `Canceled` without marking the set dirty, unlike the other cancellation paths:

```diff
     for animation in self.animations.iter_mut() {
         if animation.is_cancelled_in_new_style(new_style) {
             animation.state = AnimationState::Canceled;
+            self.dirty = true;
         }
     }
```

I confirmed this with manual tests: the original build freezes, adding this line restores all four transitions with correct intermediate and final values, and reverting it brings the freeze back.

---

**Platform:**
Arch-based Linux (CachyOS), KDE/Wayland, NVIDIA RTX 2060 SUPER.

neeopy:
Fixes servo/servo#48225

This covers the three causes discussed in the issue, all in `ElementAnimationSet::update_animations_for_new_style`:

- Canceling an animation didn't mark the set `dirty`, so the cascade never replaced the animated rule.
- Removing the last matching `@keyframes` rule while `animation-name` was still declared left the effect applying indefinitely.
- On `display:none`, cancellation happened after the rule substitution and recascade instead of before, letting a stale value resurface once the element became visible again.

Tested and validated locally: reverting each change individually brings back the corresponding symptom.


Loirooriol (commented):
Please open a companion Servo PR with test coverage

jdm:
> Please open a companion Servo PR with test coverage

That's https://github.com/servo/servo/pull/48266.

